PR #150, opened 2021-12-08
Patch 1

jurelou:
Send client events to a splunk instance

Basically, i've merged thoses artifacts together:
- https://github.com/Velocidex/velociraptor/blob/master/artifacts/definitions/Elastic/Events/Clients.yaml
- https://github.com/Velocidex/velociraptor/blob/master/artifacts/definitions/Splunk/Flows/Upload.yaml


scudette (approved):
Thanks! this is very cool!

PR #198, opened 2022-01-18
Update wordlist

mgreen27:


scudette (commented):
I actually prefer not to sort this because it makes it very hard to control which entries are added. It is better to just add the last few words to the end each time.

PR #410, opened 2022-10-21
AnyDesk artifact

DfirJos:


weslambert:
Nice work!

This caught my eye, because I had to fix some metadata around this artifact recently.

It looks like there is an existing, similar AnyDesk artifact here: 

https://docs.velociraptor.app/exchange/artifacts/pages/windows.applications.anydesk.logparser/

Would it be better to update that one to avoid duplication?

DfirJos:
Thanks. And sure, I can do that!

PR #413, opened 2022-10-25
Server Event - Add Assets to DFIR-IRIS

BadBloopZ:
Watches for clients that get the `IRIS-ADD` label assigned and adds them to DFIR-IRIS. To successfully add an asset, a label specifying the system type is required as well. See the default values of `AssetLabels`. This artifact adds information to IRIS and vice versa:
   
- As properties in IRIS: Hostname, domain, last IP address, system type
- As description in IRIS: Velo ClientId, Velo agent first seen, Time when the asset was added to IRIS
- As tags in IRIS: Existing tags for the client in Velo except for the IRIS tags and the system type labels
- As metadata in Velo: IRIS asset id

Once an asset has been added to IRIS, the `IRIS-ADD` label will be changed to `IRIS`. As long as the IRIS tag is in place, the asset won't be readded.
If it fails to add the client for whatever reason, i.e., wrong case ID, wrong API key, etc., it will change the `IRIS-ADD` label to `IRIS-ERROR`.
   
For best experience use the Exchange.Server.Monitor.Autolabeling.Clients to automatically apply labels specifying the system types to clients as soon as they are enrolled!
     
*Disclaimer: This artifact utilized some parts of Wes Lamberts Exchange.Server.Monitor.IRIS.Case.Create artifact.*

### DIFF
--- a/content/exchange/artifacts/Server.Monitor.Autolabeling.Clients.yaml
+++ b/content/exchange/artifacts/Server.Monitor.Autolabeling.Clients.yaml
@@ -1,0 +1,33 @@
+name: Server.Monitor.Autolabeling.Clients
+author: Stephan Mikiss @StephMikiss
+description: |
+    This server side event monitoring artifact watches for new client enrollments and automatically labels them according to their domain roles.
+    
+    * Standalone & Member Workstations will get the label `Workstation` assigned.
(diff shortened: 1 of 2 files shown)


BadBloopZ:
I just learned more about if statements in the discord and will update the artifact. I will reopen once the new version is ready

PR #526, opened 2023-02-19
Create Windows.Detection.PipeHunter.yaml

acedef:
If tools that use RemCom, like Impacket psexec.py, aren't cleaned up, the target is exposed to future privesc/lateral movement. Generalized the artifact enough so it could be used to scope pipe names by ACL/proc.


mgreen27:
@acedef thank you for this. If your interested you can also import the DetectRaptor content in the exchange.
This will lead to this artifact that is native VQL and has the RemCom pipe included: https://github.com/mgreen27/DetectRaptor/blob/master/vql/NamedPipes.yaml

Covers named pipes and also Sysmon for previous named piupe telemetry.
Is a bit noisy but great to stack and look for outliers.

PR #538, opened 2023-03-04
Create Windows.Registry.Bulk.ComputerName

angry-bender:


angry-bender:
Similar to the bulk registry started by to @mgreen27 I'm wondering if we keep it to a single query, or create a parameter with the CSV. 

Try to make that work before but I couldn't get two lines of inception to make it happen

Might at least make a parameter so you can change the source hive and destination key

angry-bender:
> Might at least make a parameter so you can change the source hive and destination key

Fixed in [a088a3c](https://github.com/Velocidex/velociraptor-docs/pull/538/commits/a088a3c68854f49a6caed462339090441aa40f7c)

PR #577, opened 2023-04-28
Create Windows.Forensics.Clipboard

h-adwan:


content/exchange/artifacts/Windows.Forensics.Clipboard
```diff
@@ -0,0 +1,49 @@
...
+parameters:
+ - name: FileGlob
+   default: C:\Users\*\AppData\Local\ConnectedDevicesPlatform\*\ActivitiesCache.db
+
+sources:
+  - precondition:
+      SELECT OS From info() where OS = 'windows'
+
+    query: |
+      LET FileGlob <= '''C:\Users\*\AppData\Local\ConnectedDevicesPlatform\*\ActivitiesCache.db'''
```
mgreen27, line 35: You should be able to remove this item and use the parameter definition.
As it stands the parameter is not used as its overwritten by this line.

PR #578, opened 2023-04-28
Create Windows.Forensics.Clipboard

h-adwan:


mgreen27:
Thank you for this - only one other requirement, you need to ensure your file has a .yaml extension.

PR #610, opened 2023-05-24
Artifact to add notebook rows as timeline events to DFIR-IRIS 

BadBloopZ:
This module aims to further ease the process of documenting analysis results. Using this artifact, one can comfortably add specific columns and custom text to Iris as a timeline event. IOCs and assets will be linked to correctly represent the entry and provide all the value added by cross-referencing IOCs and assets.

@weslambert I'd appreciate if you could provide feedback / ideas for improvement if you have the chance to play around with it! :)

### DIFF
--- a/content/exchange/artifacts/IRIS.Timeline.Add.yaml
+++ b/content/exchange/artifacts/IRIS.Timeline.Add.yaml
@@ -16,6 +16,7 @@ description: |
    ```
    
    - The true power of this artifact lies in the ability to quickly add many entries to DFIR-IRIS. You will **most likely use this artifact from within a notebook**.
+   - There is a basic mechanism established to stop duplicates from being added. An event is compared to existing entries based on asset name, flow id, timestamp and the description. You can add multiple events happening at the same time for the same asset originating from the same flow as long as the description varies, e.g. by including dynamic details of the activity that differentiates between the events at the same time like a process name.
    
    **Example** to add entries in a performant way and have the results in well-structured columns. **ATTENTION: ALWAYS USE ASYNC=FALSE OTHERWISE ANY ASSETS THAT NEED TO BE CREATED MIGHT BE DUPLICATED!!!**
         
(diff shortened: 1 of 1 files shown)


content/exchange/artifacts/IRIS.Timeline.Add.yaml
```diff
@@ -0,0 +1,219 @@
...
+            
+      LET metadata_preparation = SELECT client_metadata(client_id=clientId) as metadata FROM scope() WHERE metadata.IRIS_AssetId
+      
+      LET syncAsset = SELECT * FROM Artifact.Custom.Exchange.IRIS.Sync.Asset(clientId=clientId,IrisURL=URL,IrisCaseId=CaseID,IrisKey=Creds,IrisRootCA=RootCA,DisableSSLVerify=DisableSSLVerify)
+        
+      LET eventAsset1 = if(condition=metadata_preparation,then=array(a=metadata_preparation.metadata[0].IRIS_AssetId),else=if(condition=syncAsset.Result[0]="SUCCESS",then=array(a=metadata_preparation.metadata[0].IRIS_AssetId),else=[]))
+        
+      LET eventAsset2 = if(condition=AdditionalAssetId,then=split(string=AdditionalAssetId,sep=",|;"),else=[])
+        
+      LET eventCategory = if(condition=Category=~"^legit",then=2,
```
scudette, line 129: Just fyi this type of thing can now be done easier with regex_transform()

content/exchange/artifacts/IRIS.Timeline.Add.yaml
```diff
@@ -0,0 +1,219 @@
...
+                                event_color=eventColor,
+                                event_date=eventDate,
+                                event_tz="+00:00",
+                                event_content=Description,
+                                event_raw=RawEvent
+                            )
+                            ,format="json"
+                        )
+      
+      LET apiRequestIrisAddEvent = if(condition=config.version.version < "0.6.9",
```
scudette, line 174: it is better to do version camparisons on the plugin itself 

```version(plugin="http_client") < 2```

because comparing on the semantic version is not so simple

BadBloopZ: I just tested this approach, but I get version '2' for both, 0.6.8-2 and 0.6.9. Is this intended? I thought as the parameter skip_verify was added and disable_ssl_security marked as deprecated in 0.6.9, the version of the plugin should be increased as well.
image
image

content/exchange/artifacts/IRIS.Timeline.Add.yaml
BadBloopZ: Hmm, yep, most of the cascaded if statements can be easier implemented using the regex_transform function. However, there's no possibility for some sort of a final 'else' statement or a default/catch all.

weslambert:
> 

Sure thing! Sorry for the delay! I'll give it a try when I have a moment and let you know if I have any feedback 😃.

content/exchange/artifacts/IRIS.Timeline.Add.yaml
scudette: The version relates to the version of the plugin itself - in this case 2. When a plugin changes functionality in an incompatible way (e.g. adding new option) then its version is incremented independently of the release version. This makes it easier to write VQL that works with whatever release as long as the plugin needed is of the right version.

content/exchange/artifacts/IRIS.Timeline.Add.yaml
scudette: If the Category is a fixed string you can just use a lookup to sub it like this for example https://github.com/Velocidex/SQLiteHunter/blob/55607603acc357becb592d522a2430e6ed86901e/definitions/ChromiumBrowser_HistoryVisits.yaml#L55

PR #624, opened 2023-06-29
Update PrintNightmare.yaml

DakshGajjar:
According to Microsoft, this vulnerability can only be exploited if the “NoWarningNoElevationOnInstall” key in the registry is set to 1. 
   
The artifact scans the device registry to check if the beforementioned key exists or not; if it is undefined or doesn’t exist, then the system is not vulnerable to the PrintNightmare. Otherwise, the system is considered to be vulnerable to exploitation.

This vulnerability can be exploited using the Evil Printer attack.

The suggested VQL query looks for the registry values to find a registry key named “NoWarningNoElevationOnInstall”.
   
References:
   
https://msrc.microsoft.com/update-guide/vulnerability/CVE-2021-34527

### DIFF
--- a/content/exchange/artifacts/PrintNightmare.yaml
+++ b/content/exchange/artifacts/PrintNightmare.yaml
@@ -1,136 +1,52 @@
-name: Windows.Detection.PrintNightmare
-author: "Matt Green - @mgreen27"
+name: Windows.Registry.PrintNightmare
 description: |
-  This artifact returns any binaries in the Windows/spool/drivers/**
-  folders with an untrusted Authenticode entry.
+   CVE-2021-34527 or Windows Print Spooler Remote Code Execution Vulnerability
+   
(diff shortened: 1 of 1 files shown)


DakshGajjar:
Velociraptor PrintNightmare Artifact to detect the PrintNightmare variant CVE-2021-34527.

DakshGajjar (commented):
.

DakshGajjar (commented):
.

scudette:
This looks like an additional but different detection technique to the original artifact.

Maybe add a second source to the previous artifact or make a new one?

DakshGajjar (commented):
.

DakshGajjar:
> This looks like an additional but different detection technique to the original artifact.
> 
> Maybe add a second source to the previous artifact or make a new one?

sure, but it is showing some error regarding SpellCheck.

DakshGajjar:
I have created a new one.

DakshGajjar:
It's waiting for approval, so I have created a new one, is that okay or
should I continue with the existing one?.

Waiting for your reply.

Best Regards,
Daksh

scudette:
No it's fine we can add a new one

PR #630, opened 2023-07-10
Create Windows.Detect.Steam.on.Client

MartvandeLogt:
An Artifact to detect Steam on a Windows Client.


MartvandeLogt:
Closed because I found the Artifact to actually look for a specific file.

PR #636, opened 2023-07-25
Create Windows.Applications.Chrome.Bookmarks.yaml

sikhamohan:
Artifact to fetch bookmarks from Windows Chrome browser


scudette:
Just took at a look at this file and it looks like it should be recursive. It looks like it is possible to create a folder inside each folder and put the bookmark in it. The code really needs to be able to recurse into the folders to extract each bookmark

image

scudette:

PR #682, opened 2023-09-18
Fixes wrong path variable in Bitsadmin artifact

f-block:
FullPath is not defined/part of any surrounding select statement and hence not known to the query. This artifact fails with:

```
ERROR: Symbol FullPath not found. Current Scope is: ...
```


scudette (commented):
FullPath is actually provided from the files query. In recent velociraptor versions we have deprecated FullPath in favour of OSPath so probably we need to change that instead

f-block:
Interesting, using `OSPath` does indeed fix it, even though other artifacts still work with `FullPath` and just note the deprecation. Probably depends on which plugin/function is used.

PR #702, opened 2023-10-07
Add an artifact that queries connections config from NetworkManager

misje:
NetworkManager is commonly used in Linux for configuring network connections. This artifact extracts all network connections configured in NetworkManager, which includes data like

- Network names
- SSIDs
- BSSIDs,
- Wi-Fi PSKs (redacted by default)
- Static IP address assignments
- Routes

If NetworkManager is used on the system, this gives information about what networks the computer has connected to (especially for Wi-Fi), and when it was last connected to those networks. A simple VQL filter reveals which of those network were insecure, or if any interfaces has been configured as an access point or as ad-hoc. NetworkManager also tracks all seen BSSIDs for networks.

Connections are typically not tied to a specific interface, so it is not always useful to correlate current IP addresses and routes on interfaces against active connections. An optional enrichment using interfaces() was tested and subsequently removed.

INI is a very common configuration file format, and I couldn't find any existing parser. This artifact therefore exports an INI file parser for use by other artifacts. `parse_ini(filename)` returns three columns: Section, Key, Value. `parse_ini_as_dict(filename)` returns a single column, Contents, with everything in a dict. Section names are prepended to keys, separated by a slash. Spaces in values are supported, and quotes are included as-is.

I haven't included any example output because there would be little data left after sanitising everything.

### DIFF
--- a/content/exchange/artifacts/Linux.Network.NM.Connections.yaml
+++ b/content/exchange/artifacts/Linux.Network.NM.Connections.yaml
@@ -14,15 +14,14 @@ description: |
   is the default). Note that by default, NetworkManager doesn't managed
   connections defined in /etc/network/interfaces.
 
-  Information such as IP addresses, routes and other interface settings
-  and statistics are not retrieved. This information may be retrieved
-  from Generic.Network.InterfaceAddresses or interfaces(), but note that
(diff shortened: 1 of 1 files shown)


misje:
I have a few other features to commit, and I'll also provide some example output.

PR #706, opened 2023-10-13
Create IPCheck.Virustotal

AdrianX21:
I want to provide this artifact so that analysts can verify IP from the notebook. In my case it has been very useful to investigate the connections established on the machines and look for possible C2C and data exfiltration.


AdrianX21 (commented):
image

PR #729, opened 2023-11-07
Create Windows.Sys.BitLocker.yaml

zaneGittins:
Get BitLocker volumes, including keys, using PowerShell.


zaneGittins (commented):
Cleaned up some of the PowerShell.

PR #767, opened 2024-01-20
Enhacement Windows.Forensics.PersistenceSniper Artifact 

ablescia:
I implemented a new version of the PersistenceSniper Artifact:

Below are the improvements:

- Memory avoiding usage to store the PersistenceSniper results. Now the artifact uses a csv file;
- Support for **-IncludeHighFalsePositivesChecks** PersistenceSniper argument
- Support for **-DiffCSV** PersistenceSniper argument
- Avoid nested PowerShell calling
- VQL code optimization and readability
- Support scan result upload


mgreen27 (approved):
looks good - thankyou for submitting the PR

mgreen27:
The only change I would recommend is remove the 2 scopes and automatic update.
I for one prefer to not upload a csv as the data is collected in the actual query and your collecting twice. This can add up when running in a large network.

If you think you need an upload, consider adding as a boolean value and implement similar to this: https://github.com/Velocidex/velociraptor/blob/ca10bb821d8630fce21e9095ea4ca2f8478de863/artifacts/definitions/Generic/Detection/Yara/Glob.yaml#L139-L145

ablescia:
Dear @mgreen27,
in some situations, it is useful to send the csv file to the server because it might contain new elements to be inserted in the file **false_positives.csv** without manually deleting the columns **FlowId**, **ClientId**, and **Fqdn**.
I will proceed with the modification you suggested.

ablescia:
Dear @mgreen27 in my last commit, you can find the UploadHits flag implementation as you suggested.
Best Regards,

mgreen27 (approved):

mgreen27:
@ablescia FYI - I think you can collect this csv much easier by using gui and deselecting columns and exporting.
On a single machine the upload method is ok, but running across many machines, this method allows you to extract the results from a hunt much much eaiser.
image

If the download is something you will use I will approve.

ablescia:
Dear @mgreen27 ,
Yes, I am currently using the exporting function that you mentioned and it works!
Thank you.

PR #807, opened 2024-03-05
Create Generic.Enrichment.ThreatFoxHunt.yaml

Director-Fusion:
Use threatfox to check logs for records of connections to IOCs on threatfox.


Director-Fusion:
What do you need from me to get this in?

PR #872, opened 2024-07-06
Update IRIS timestamp formatting logic

predictiple:


content/exchange/artifacts/IRIS.Timeline.Add.yaml
```diff
@@ -171,13 +171,11 @@ sources:
...
-                                          if(condition=Timestamp.Nanosecond, then=
-                                               parse_string_with_regex(
-                                               string=str(str=Timestamp.Nanosecond),
-                                               regex='''(^\d\d\d)''').g1
-                                               ,else="000")
-                                          ])
+      LET eventDate = format(format="%d-%02d-%02dT%02d:%02d:%02d.%03.f", args=[
+                                      Timestamp.Year, Timestamp.Month, Timestamp.Day,
+                                      Timestamp.Hour, Timestamp.Minute, Timestamp.Second,
+                                      Timestamp.Nanosecond / 1000000
```
scudette, line 186: Are you sure this is correct? My understanding is that nanosecond is the total time in the minute in nanoseconds so it already includes the second part. Its not just the fractional part of the second.

predictiple: It seems to be only the fractional part

Screenshot from 2024-07-06 13-08-35

https://pkg.go.dev/time#Time.Nanosecond

mgreen27 (approved):

PR #875, opened 2024-07-10
Add artifacts to monitor network communication on clients

sec-hbaer:
Hi all,

we created artifacts to complement Process Tracking for Windows with network connection tracking.
The artifact was tested on Windows and Linux. Due to using built-in functions only, it should work on all supported OSs.

network connections are tracked by periodically diffing the netstat info. Added or removed connections are identified by combining Pid, layer 3 protocol, layer 4 protocol, local address used, local port used, remote address used, remote address used. Based on the PID, network information is enriched with process information.

The client monitoring is complemented with a server alerting artifact. That artifact allows sending messages via webhook about the tracked network connections. Connections can be filtered on the client that emitted the connection, the name of the process that started the network connection, the remote ip being contacted and the remote port being contacted. Those filters should also be used, as it is easy to spam the webhook (lots of network activity by default :D)

There will be a blog post by SEC Consult soon about our though process behind creating these artifacts and some recommendations based on our tests.

I wish the community much fun with this contribution. :)

### DIFF
--- a/content/exchange/artifacts/Server.Alerts.TrackNetworkConnections.yaml
+++ b/content/exchange/artifacts/Server.Alerts.TrackNetworkConnections.yaml
@@ -30,7 +30,7 @@ sources:
     - query: |
         SELECT * FROM foreach(
           row={
-            SELECT *, client_info(client_id=ClientId).os_info.fqdn AS Fqdn from watch_monitoring(artifact='SEC.N01_Preparation.TrackNetworkConnections')
+            SELECT *, client_info(client_id=ClientId).os_info.fqdn AS Fqdn from watch_monitoring(artifact='Exchange.Generic.Events.TrackNetworkConnections')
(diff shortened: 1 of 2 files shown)


content/exchange/artifacts/Server.Alerts.TrackNetworkConnections.yaml
```diff
@@ -0,0 +1,46 @@
...
+    description: Regex for filtering on the remote ip connected to
+  - name: RemotePortRegex
+    type: regex
+    description: Regex for filtering on the remote port connected to
+
+sources:
+    - query: |
+        SELECT * FROM foreach(
+          row={
+            SELECT *, client_info(client_id=ClientId).os_info.fqdn AS Fqdn from watch_monitoring(artifact='SEC.N01_Preparation.TrackNetworkConnections')
```
mgreen27: you might need to modify this line to match your monitored artifact: Exchange.Generic.Events.TrackNetworkConnections

Note the "Exchange." prefix as this is added on the typical import fro the artifact exchange.

sec-hbaer: true, thanks for catching that :D

mgreen27 (approved):

PR #882, opened 2024-07-16
Add a powerful util for installing deb packages

misje:
In Debian-based OSes software is preferably installed through the package manager. This artifact provides a comfortable way to install packages, as well as configuring them using debconf. The package can either be a name or a deb file, which can be optionally uploaded.

Other artifacts can offload all installation and configuration logic to this artifact by calling it like this: `SELECT * FROM Artifact.Linux.Utils.InstallDeb(DebName='sl')` or `SELECT * FROM Artifact.Linux.Utils.InstallDeb(DebName='/tmp/mypackage_0.1.0-1_amd64.deb')`. A row with *Step* "Installing package" and *ReturnCode* 0 will indicate that the installation was a success.

### DIFF
--- a/content/exchange/artifacts/Linux.Utils.InstallDeb.yaml
+++ b/content/exchange/artifacts/Linux.Utils.InstallDeb.yaml
@@ -13,12 +13,14 @@ reference:
 parameters:
    - name: DebName
      description: |
-        Package to install (by name). Ignored if DebFile is set.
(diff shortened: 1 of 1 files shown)


misje:
Perhaps this should go in the main repo instead, otherwise integrated artifacts cannot use it (they shouldn't depend on artifacts from the exchange)? Tell me if that sounds alright, and I'll create a new pull request.

PR #883, opened 2024-07-17
Fix Notebooks.Admin.Flows artifact

predictiple:
Something weird in the original version is causing [this part of the code](https://github.com/Velocidex/velociraptor/blob/b4e45a3a4b09268a877c6270b16c0ec95884bbe3/services/repository/manager.go#L190) to trigger, which means that artifact_set() fails.

All I did in this change is to create a new yaml file and copy over the content. Then delete the original file and give it's name to the new file.

The artifact import now works.


scudette:
Good catch! There is some binary before the `name:` line

image

scudette:
Looks like its a BOM added by visual studio
https://stackoverflow.com/questions/44098326/ef-bb-bf-at-the-beginning-of-json-files-created-in-visual-studio

PR #903, opened 2024-08-30
added artifact for parsing FreeBSD utx files

Herbert-Karl:
Hi, 

I wrote an artifact to parse the utx file under FreeBSD. 
The file is similar to the wtmp file under Linux. It records logins, logouts, boots, shutdowns and system time changes.

I appended the results from running the artifact on a test system (freebsd 14.1-RELEASE-p3 amd64)
[example_results.json](https://github.com/user-attachments/files/16818996/example_results.json)

Im not sure, if this artifact is better of in the Artifact Exchange or shipped by default with Velociraptor. For now, I offer it here.
If there is potential to improve the VQL, please comment so :)

Kind Regards

### DIFF
--- a/content/exchange/artifacts/FreeBSD.Sys.Utx.yaml
+++ b/content/exchange/artifacts/FreeBSD.Sys.Utx.yaml
@@ -1,0 +1,76 @@
+name: FreeBSD.Sys.Utx
+author: Herbert Bärschneider
+description: |
+  Parse the utx file of the system (similar to wtmp on Linux). This covers user sessions, boots, shutdowns and system time changes.
+  Because FreeBSD discards fields for the entries of the utx file based on type (see `man getutxent`), no direct parsing of the file using "vtypes" is done (too complicated for me to define a structure for parsing), but rather native tools are used for accessing the data.
+  
+  Using a value of "time" with the "userRegex" Parameter will give you all entries related to boots, shutdowns and system time changes.
+  
+  Beware: logout times are given in localtime! Furthermore, that column is not automatically parsed into timestamp values, because the tool output is not consistently a recognizable datetime.
+
+type: CLIENT
+
+parameters:
+  - name: utxGlobs
+    default: /var/log/utx.log*
+    description: |
+      glob for covering the files that should be parsed; default covers the usual location of the utx file on FreeBSD
+  - name: userRegex
+    type: regex
+    default: "."
+    description: |
+      Regex for filtering the users, showing those you are interested in 
+  - name: DateAfter
+    type: timestamp
+    description: |
+      timestamp used for filtering the login time
+  - name: DateBefore
+    type: timestamp
+    description: |
+      timestamp used for filtering the login time
+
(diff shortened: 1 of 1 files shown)


content/exchange/artifacts/FreeBSD.Sys.Utx.yaml
```diff
@@ -0,0 +1,76 @@
...
+      })
+      
+      -- parse the output from each file as json
+      LET UtxContent = SELECT parse_json(data=Stdout) as Content FROM UtxParsing
+      
+      -- pull out the interesting information from the nested json content
+      LET FormatedContent = SELECT * FROM foreach(row=UtxContent,
+      query={
+        -- the if-blocks are added because some Symbols are not present in each row, creating ugly errors in the flow if naively selected
+        SELECT user, timestamp(string=`login-time`) AS login_time, if(condition={SELECT * FROM scope() WHERE `logout-time`},then=`logout-time`,else="") AS logout_time, if(condition={SELECT * FROM scope() WHERE `session-length`},then=`session-length`,else="") AS session_length, if(condition={SELECT * FROM scope() WHERE tty},then=tty,else="") AS tty, if(condition={SELECT * FROM scope() WHERE `from`},then=`from`,else="") AS remote, Content.`last-information`.utxdb AS utx_log FROM foreach(row=Content.`last-information`.last)
```
scudette: there is no point having 
```
if(condition={ SELECT * FROM scope() where `logout-time`, then=`logout-time`, else="")
```

this is just the same 

```
`logout-time` || "" 
```

Herbert-Karl: I'll have to test that and will report back, if I can simplify that part of the query.
I created the construct to solve the "problem" that some entries in `Content."last-information".last` do not have an element "logout-time". This would create errors, which showed the flow in the UI as having errored (which I thought might confuse people using the artifact).

scudette: not a problem - usually you can suppress the symbol not found errors using something like

```sql
LET s = scope()

SELECT ... s.logout_time AS logout_time ....
FROM ....
```

This works because s is the scope dict and s.logout_time means to pull out the logout_time member in that dict- so if it is not there the default is just NULL not a symbol not found error.

Also there is no need to surround logout_time in backticks because its name does not contain any VQL operators (like for e.g. space or dots)

Herbert-Karl: ~~The two statements seem not to be the same. I changed the query as you suggested, but now I get following errors, which mark the flow as having errored:~~
image

~~I rather prefer having the flow not show as errored, as this might confuse an analyst using it.~~
I need to retest, didnt see your last message :D

Herbert-Karl: with your last suggestion, it works and the query looks a lot nicer.
Thanks :)

PR #913, opened 2024-09-18
Add image guidelines doc

predictiple:


dev/image-guidelines.md
```diff
@@ -0,0 +1,329 @@
...
+
+While screenshots do need to depict the actual application UI, it's reasonable to tweak some aspects to make the screenshots a bit clearer for documentation purposes.
+These may include:
+1. using a slightly heavier font weight: 500 rather than 400. On Windows the fonts seem to render slightly lighter for some unknown reason.
+2. using pure black or white as the text color instead of lower contrast tones.
+3. reducing the opacity of the modal backdrop so that items in the background are not excessively dimmed
+4. deleting distracting elements from the DOM if necessary.
+5. adding a glow effect via CSS to emphasize a specific element.
+
+Points 1-3 can be patched into the GUI's CSS and compiled to make a documentation-specific binary. A patch is provided in this folder named `documentation_tweaks_{commit_id}.patch`
```
scudette: Can we just add this as another theme?

dev/image-guidelines.md
```diff
@@ -0,0 +1,329 @@
...
+For annotated screenshots this aspect contrasts well with inorganic GUI elements, and in general it's humanistic and visually appealing.
+
+For **unannotated screenshots** we use the PNG format since it has lossless compression which preserves the legibility in text in images of the Velociraptor GUI. There's no need to do anything fancy with these images. Just try to keep the size below 300KB if possible.
+
+## Fonts
+
+The standard Excalidraw version uses a default font named Virgil which is too scrawly and therefore a bit hard to read (as discussed [here](https://github.com/excalidraw/excalidraw/issues/2945)).
+The default fonts are not able to be changed without editing the application's source, and the Excalidraw developers have stated disinterest in make custom fonts possible.
+
+Fortunately there exists a reasonably well-maintained fork which allows adding custom fonts
```
scudette: Im a bit reluctant to use an unofficial fork - do we absolutely need it? Dont we replace the fonts anyway?

The convenience with the official project is that there is a free site we can use without needing to mess with yarn - this makes it a lot more accessible.

dev/image-guidelines.md
```diff
@@ -0,0 +1,329 @@
...
+:   https://github.com/hulkbig/excalidraw-zh
+
+It's very easy to set up a local instance:
+
+1. clone the repo and cd to it
+2. `yarn` to install
+3. `yarn start` to start the web app
+4. follow the link that it shows you (by default http://localhost:3000/).
+
+Only 3 font variants are supported so we assign them as follows:
```
scudette: I like the idea of a consistent font everywhere - lets just coerce it with a presubmit - that way this docs are not relevant

dev/image-guidelines.md
```diff
@@ -0,0 +1,329 @@
+# Guidelines for creating image content
```
scudette, line 1: We need to add a hugo preamble here to render

dev/image-guidelines.md
```diff
@@ -0,0 +1,329 @@
...
+## Exporting images
+
+### Formats and settings
+
+When exporting images from Excalidraw we include "Background" which encodes the screenshot (if there is one) plus a small margin around it. For line diagrams this option creates an opaque white background plus a small margin.
+
+The Scale setting is a bit confusing:
+
+- If you've inserted a screenshot and not resized it, then export it to PNG with Scale=1x, this will produce an image that is about 1/2 the size of the original. For that reason either resize the image on the canvas to approximately it's original size or else choosing Scale=2x for the export will produce an approximate 1:1 of the original size. In other words the Scale setting is relative to the size of the image on your screen.
+- For SVG exports the Scale setting is irrelevant: it will base64-encode the screenshot at it's original size, even if you've scaled the screenshot on the Excalidraw canvas.
```
scudette: Lets just standardise on svg - there is no point messing with scaling here and we want to reduce decision friction

scudette:
I like having a style guide for documentation to ensure consistency. However rather than being indecisive and introduce more options, we should be totally prescriptive and just pick a certain option each time.

See similar style guides https://google.github.io/styleguide/go/ and how they are phrased - a decision then a rational is ok but the less wiggle room there is the better. Choices add friction and reduce consistency for the sake of some aesthetic and subjective  freedom .

dev/image-guidelines.md
predictiple: We could but then it probably shouldn't be one that's visible/selectable for users. If there's some way to hide it in the GUI user prefs but still allow it to be set it via user_options() that would be ideal. It seems like overkill though.

Since those 2 CSS files hardly ever change it's easy to just patch it when needing to make screenshots. And updating the patch on rare occasions isn't difficult.

dev/image-guidelines.md
predictiple: The official version doesn't allow custom fonts at all. That means you can't see an accurate representation of what you're working on while you're working on it.

All the fonts are also variable width so there will definitely be significant differences if we swap a font with a _different_ font, especially given that the official version's default "Virgil" is a very wide font. When we set the font URL in the SVG we're actually undoing the swap-to-defaults that the SVG export performs.

Looking at the SVGs that you recently created, which I assume came from the official hosted app, they have one font referenced:
Screenshot from 2024-09-18 16-39-56
There's no indication that the embedded font there is actually "Virgil".

I've just done a quick test now at excalidraw.com: added text in one of each of their 3 fonts, then exported to SVG. The file contains an embedded "Excalifont", _5 embedded copies_ of "Nunito", and "Comic Shanns" (which is supposed to be their code font but looks not at all code-like). Until now their fonts have always been Virgil (now rebranded to Excalifont), Cascadia, and Assistant.

So problem is really the opposite, where using the hosted app means they can change things on a whim and and offer no guarantees of backward compatibility. The fork doesn't have any issues - all the functionality works - and can always pin a version if necessary. I don't think there will be any new features in the official version that we will be yearning for - the app does what's needed.

dev/image-guidelines.md
predictiple: This doc is not intended to be published via Hugo, at least not any time soon. It's just developer notes, similar to the ones in the velociraptor repo that explain GUI development and translations.

At some point it might form part of a contributor guide once we figure out how to present that.

dev/image-guidelines.md
predictiple: Yes agreed. I think that's what I meant to say but wrote that before I had come to that realisation.

dev/image-guidelines.md
scudette: I understand the layout can differ but is it really that important?

I generally prefer things that take least effort for the person developing the doc and leave more of the look and feel to the site builder scripts - that way there can be less variability and the user only needs to think about what to write not how to lay the text out.  In other words it does not matter if the text is slightly bigger or smaller as long as the bounding box is set over the correct region.

dev/image-guidelines.md
scudette: We should publish it - maybe as a knowledge base item. 

If we dont intend people to read it then whats the point of writing it?

dev/image-guidelines.md
scudette: I dont mind it being selectable - that makes it easier for people to use. I dont think it is acceptable to patch the code just for docs - this is likely to result in accidents like commiting this to the main branch with another PR accidentally.

dev/image-guidelines.md
predictiple: We have to do the font link substitution regardless of how we do it. We swap the links to keep the fonts the same.

In the app you have to configure the fonts so that you can see them while you are drawing. As I mentioned in another comment the various fonts are variable width so a blind swap will cause unexpected shifts in their positioning

scudette: Yeah the font will be repositioned within the bounding box - but as long as the bounding box is reasonable it should flow within it.

Just tried it and it makes no practical difference to the font positioning:

Original font
image

Replaced with our font
image

I lean towards just making the instructions simpler. If a power user really wants to use the exact same font - go for it. But we dont need to mandate that in this style guide.

predictiple:
> I like having a style guide for documentation to ensure consistency. However rather than being indecisive and introduce more options, we should be totally prescriptive and just pick a certain option each time.

I've reduced the wiggle room as much as I think is possible. There are many things that we can't be absolute on because they depend on the specifics of the situation and we can't anticipate and enumerate all situations. For example with the numbered annotations it depends on how obvious the particular screenshot will be to the average user. Some might need explanations attached while other not. Some might need arrows to spatially separate numbers that would otherwise cluster together. Since it's drawings there unavoidably has to be some artistic wiggle room.

As we go we are going to be discovering what works best in specific situations, or decide that things we once thought were a good idea were maybe not. So the doc is a "living document" rather than a syntax specification. It's a compilation of notes on what we _currently_ think is a good, but mainly untested, idea.

dev/image-guidelines.md
```diff
@@ -0,0 +1,335 @@
...
+Points 1-3 can be patched into the GUI's CSS and compiled to make a documentation-specific binary. A patch is provided in this folder named `documentation_tweaks_{commit_id}.patch`
+
+A glow effect (point 5) around GUI elements can be achieved by applying this CSS to the DOM elements in "dev tools".
+
+Background color may be turned off if it isn't needed or produces a bad result. It's purpose is to fill in rounded corners behind the element being highlighted.
+
+Z-index is only sometimes needed for elements such as buttons when an adjacent element has a z-index property set that causes it to overlay the glow. This property ensures that the glow effect is at the top.
+
+```css
+box-shadow: 0px 0px 10px 5px rgba(34,139,230,0.9);
```
scudette: Although I think it looks cool I dont think it works too well on all elements and style:

With the background-color it doesnt quite look like your screenshot
image

Without the background color it looks closer but messes up some icons
image

We also need to be careful to not place the highlight on the td as it basically messes it right up
image
The box shadow does not work on td and only the background color covers the whole thing.

I am also a bit concerned with the level of effort this requires to update screenshots - if we do the highlighting in the drawing program it is just a matter of replacing the underlying png layer and aligning with existing arrows / text,  but now we have to go through all the elements to re-highlight them before we take the screenshot which is more work.

Simple circles and squares are probably good enough and take less effort 
image

dev/image-guidelines.md
```diff
@@ -0,0 +1,335 @@
...
+
+These are explained below.
+
+#### Partial screens
+
+Screenshots that show only part of the screen area are relatively straightforward.
+
+![](image-guidelines/screen_partial.png)
+
+A useful trick in Chrome Dev Tools is that you can right-click on a DOM element and capture an image of just that element.
```
scudette: I tried this technique but it is less than satisfactory . The first issue is that chrome saves the file instead of copy to the clipboard meaning i have to double click it open in another program then copy paste into excalidraw increasing the time and complexity.

Then I really want to show something like 
image

But there is no html node that directly corresponds to - the closest thing is the top pane
image

but you can see the margins are very tight so its not a good screenshot. 

If I need to crop it anyway there is no point in using chrome to take the screenshot.

scudette: I think we should standardise on the framing of the screenshot though - usually we should use the full headers and sidebar because this gives navigation context as to where the user is in the app. So I do think it is important to think about this.

dev/image-guidelines.md
```diff
@@ -0,0 +1,335 @@
...
+
+#### Full screens
+
+When full screenshots are needed we can improve their legibility by simulating a smaller screen size.
+The "dev tools" of modern browsers allows you to simulate arbitrary screen sizes, including many common phone screens.
+Velociraptor's GUI is not a fully responsive web application but it's layout remains usable at quite small (by modern standards) screen resolutions.
+
+For full screenshots use one of the following 2 sizes:
+
+- 1280x960 (preferred)
```
scudette, line 151: This is very important - we do need to be consistent with the sizes

dev/image-guidelines.md
```diff
@@ -0,0 +1,335 @@
...
+If the image is reused in multiple places on the docs website (or other Velociraptor repos such as presentations) we only need to keep *one instance* of the `.excalidraw` file.
+
+
+> [!NOTE]
+> Regardless of the export format, remember to also save a copy of the drawing in Excalidraw format so that others can edit it in future, if necessary.
+>
+
+Annotated screenshots exported from Excalidraw as SVG images contain a base64-encoded copy of the screenshot. The saved `.excalidraw` file does too. So it is not necessary to keep a copy of the original screenshot PNG.
+
+### Fonts in SVG images
```
scudette, line 250: This section should just say - fonts are normalized to the standard fonts used throughout the site so dont worry about specific fonts in your drawings.

The technical details are not relevant and could change in future - also not really actionable by readers.

dev/image-guidelines/excalidraw_default_fonts.png
scudette, line 1: this is ok to show just as an FYI but consider dropping it because its not really relevant to the reader

dev/image-guidelines.md
misje: Have you tried «Take screenshot» in Firefox, in the right-click menu? It lets you choose a document element and snap the borders to that, but you may still override it later by, say extending it. It doesn't come with drawing tools, however. It saves both to file and the clipboard.

dev/image-guidelines.md
predictiple: Ok. Thanks for adding it: https://github.com/Velocidex/velociraptor/pull/3764

dev/image-guidelines.md
predictiple: The point of writing it - at this time - is so that you as I can agree on basic conventions. Basically I'm saying "This is how I do it. These are the things I think about when doing it. And can we both do it this way, or would that cause problems? Etc."

The idea of having the "dev" folder is that it's a place to share notes and supporting files, similar to the "scripts" folder but for things that aren't scripts.

The reason for making it markdown is so that it can be viewed with images and formatting: either on Github or locally. It's by coincidence that the website docs are also written in markdown.

Eventually it will be great to have a proper development guide [something like this](https://docs.excalidraw.com/docs/introduction/development) and it would be it's own section on the docs site. This could be pulled into that along with other docs from the velociraptor repo, but that is going to involve proper planning and probably a lot more explanatory writing.

dev/image-guidelines.md
predictiple: Are you saying 1280x960 should be an absolute requirement? It should be fine in most cases but I'm sure there are cases where it's not, like this one from your detection engineering post for example:
live_sigma_detection
Discretion seems necessary.

scudette: not a hard requirement - Im just saying I agree with you specifying it as preferred - it is an important point! This is great!

dev/image-guidelines.md
predictiple: > I think we should standardise on the framing of the screenshot though - usually we should use the full headers and sidebar because this gives navigation context as to where the user is in the app. So I do think it is important to think about this.

Yes, when screenshotting only a part of the screen then it's important to anchor it by including some familiar element like part of the header or sidebar. Unless a previous screenshot has already provided that spacial/navigational context.

predictiple: > I tried this technique but it is less than satisfactory . The first issue is that chrome saves the file instead of copy to the clipboard meaning i have to double click it open in another program then copy paste into excalidraw increasing the time and complexity.
> 
> Then I really want to show something like image
> 
> But there is no html node that directly corresponds to - the closest thing is the top pane image
> 
> but you can see the margins are very tight so its not a good screenshot.
> 
> If I need to crop it anyway there is no point in using chrome to take the screenshot.

All valid points. It won't work in all situations but in many situations it's more precise than capturing a screen area by manually doing a rectangle-select with some external tool. With a rectangle-select it's easy to accidentally include edges of adjacent elements or even things that border your virtual screen. That appears to have happened in your first screenshot above which has a black border.

I should have been cleared that it's provided as a tip/trick which works well in certain circumstances. It's definitely not a mandated technique. I'll annotate it as such.

The tight margins are actually a good thing because Excalidraw is going to add margins.

dev/image-guidelines.md
predictiple: The instructions are not going to be simpler if you have to explain that the user should imagine different fonts while drawing, other than the ones their eyes are seeing.

query_api_client

I've done this test with Excalidraw (the official version) and the bounding box resizes vertically and wraps the text. So that would require us to always add extra space to the bounding box so that it doesn't change size with the font swap. And then also try to explain why that's necessary and provide some estimation of how much extra space is sufficient.

Screenshot from 2024-09-19 11-21-30

I would really like to use the official version (although not their hosted app which changes without notice) but they don't support custom fonts and their default fonts suck. If you read their issues board it's frustration that a lot of people have. They've added a font picker which hints at custom font becoming possible in future, but not yet: https://github.com/excalidraw/excalidraw/pull/8012

The official version is currently doing this with the SVG exports:
- allowing 4 default fonts (3 GUI presets + 1 extra)
- embedding them multiple times

Screenshot from 2024-09-19 11-44-07

We could swap a 4-font definitions section and map to those font names but what if they decide to add more or change the names of the existing ones? They're done that in the past (where "Virgil" became "Excalifont") and the change I linked to above suggests that the font aspect is currently very much in flux and undecided.

dev/image-guidelines/excalidraw_default_fonts.png
predictiple: I've added some extra explanation. I think it's relevant for the reader to visually comprehend the substitution.

dev/image-guidelines.md
predictiple: If the bounding box is configured with extra room for text expansion that partly mitigates the problem. That is IF the user constantly remembers to do this for every bounding box.

Even so there are situations where it doesn't work:

1. If text is left-aligned but manually positioned to be centered under another object, then the font changes size and the text extended to the right and possibly even wraps. It leaves an unbalanced result that wasn't the author's intention.
Screenshot from 2024-09-19 12-27-14

2. If the text is associated with an object on it's right and the text is not set to right-align, then swapping can cause it to overlap the object on the right or shrink away for the object leaving a big gap. So that means every time you have text associated with something on it's right you have to remember to make the text right aligned and extend the bounding box to the left.
Screenshot from 2024-09-19 12-32-28

Keeping all these possibilities in mind sounds like much more effort to me. I never have to worry about whether the bounding boxes will be sufficient or if I've left some text left-aligned when it properly should be right-aligned. WYSIWYG is really really important.

dev/image-guidelines.md
predictiple: Sure this was just an idea. I've updated the doc to use drawn shapes instead.

dev/image-guidelines.md
predictiple: Sure once we've settled on a process that works and the specifics of that comes down to these choices:

1. Use Excalidraw.com hosted app. Use their fonts, use carefully designed bounding boxes and alignments, and accept that quite often the SVGs could be significantly messed up after the font swap, something which we'll only know about once the images have been published to the docs site. Also expect breakages without warning as the online app evolves.

2. Use a locally installed instance of the official Excalidraw. That eliminates the unexpected breakages but otherwise has all the same downsides as the 1st option.

3. Use a locally installed version of the excalidraw-zh fork, which allows WYSIWYG editing and generates SVGs with a 1-to-1 font spec that can be swapped without unexpected consequences. It provides all the desired functionality that the official version offers, or even promises to offer in future. It's works and it's not going to suddenly break.

scudette: No problem - it is actually not that bad because usually when writing docs we have to run hugo anyway so the font substitution happens in real time and you can see updates to the svg immediately.

predictiple: Ok, we still need to pick one though because they all embed different font names in the vector part and we just want to replace the font definition part to achieve the right mapping.

predictiple: Just added a commit that shows the font defs swap we need to make when exporting SVG from excalidraw.com

PR #931, opened 2024-10-22
ESET Endpoint Security Log Parsing Artifact

j91321:
Artifact that uses ESET Log Collector tool to convert threat logs into XMLs and parse them through Velociraptor. Covers all important log sources that are commonly encountered in enterprise setting. Tested on ESET Endpoint Security. 

Examples:
image

image

image

### DIFF
--- a/content/exchange/artifacts/ESETLogs.yaml
+++ b/content/exchange/artifacts/ESETLogs.yaml
@@ -1,0 +1,69 @@
+name: Windows.ESET.Logs
+description: |
+   Parse logs from ESET antivirus products. This log contains information about detections made by the ESET modules such as Real-time filesystem proteciton, Firewall, HIPS, Device Control, HTTP filter, AMSI Scanner etc.
+ 
+type: CLIENT
+ 
+tools:
+  - name: ESETLogCollector
+    url: https://download.eset.com/com/eset/tools/diagnosis/log_collector/latest/esetlogcollector.exe
+    serve_locally: true
+ 
+precondition: SELECT OS From info() where OS = 'windows'
+ 
+parameters:
+   - name: LogTargets
+     default: "Threat,Hips,Fw,Web,Dev,BlkF"
+     description: Selection of log targets to collect
+ 
+required_permissions:
+- EXECVE
+ 
+sources:
+  - name: Antivirus
+    query: |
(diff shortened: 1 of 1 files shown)


content/exchange/artifacts/ESETLogs.yaml
```diff
@@ -0,0 +1,70 @@
+name: Windows.ESET.Logs
+author: Ján Trenčanský - j91321@infosec.exchange
+description: |
+   Parse logs from ESET antivirus products. This log contains information about detections made by the ESET modules such as Real-time filesystem proteciton, Firewall, HIPS, Device Control, HTTP filter, AMSI Scanner etc.
+ 
+type: CLIENT
+ 
+tools:
+  - name: ESETLogCollector
```
mgreen27, line 9: Can you add an expected_hash field in?

tools:
  - name: ESETLogCollector
    url: https://download.eset.com/com/eset/tools/diagnosis/log_collector/latest/esetlogcollector.exe
    expected_hash: efade9e0e6d228c118c4c28de2e45c37ae2fa4973283d5f851e6c62c7b72f53b < add the hash here
    serve_locally: true

mgreen27: im also wondering what this tool does -  for example: does it decode the logs?
Depending on what it does we might be able to do this natively and bypass the need for EXECVE

j91321: I've added expected_hash. The main purpose of ELC here is that it can convert proprietary format of AV logs into XML. See https://help.eset.com/elc/4.10/en-US/elc_cli.html It's too bad AV can't currently convert logs to XML directly without the need to download this tool.

mgreen27: no worries - this is defiantly a useful artifact!

I just like to mention native as Ive seen logs simply rc4-ed that were easy to do natively so think about uplifting this if we can figure out what they are. VR can definitely read XML, just what ever the other binary format is. The general idea of minimising EXECVE and third party execution on the endpoint is always a good one. :)

j91321: Agree, I will look into if anything can be done about it. I'm probably best person to do it as I work for ESET.

PR #984, opened 2025-01-23
Update Windows.Audit.CISCat_Lite artifact

utkutombul:
Windows.Audit.CISCat_Lite artifact has a problematic `execve` command which doesn't properly execute in cases where CISCatPath has spaces in file path. I solved the problem by executing via PowerShell artifact instead of `execve`, it properly executes now.

### DIFF
--- a/content/exchange/artifacts/Windows.Audit.CISCat_Lite.yaml
+++ b/content/exchange/artifacts/Windows.Audit.CISCat_Lite.yaml
@@ -43,7 +43,8 @@ sources:
           
       -- Run the command     
       LET _ <= SELECT *
-        FROM execve(argv=[CISCatPath, "-b", BaselinePath, '-p', ProfileName])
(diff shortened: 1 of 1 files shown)


scudette:
This is actually very interesting because execve is supposed to handle spaces in paths by itself. 


We usually try to avoid running PowerShell unnecessarily because it pollutes the PowerShell artifacts.

scudette:
Its definitely not an issue of the spaces in paths
image

Maybe there is something else about the powershell environment that makes it work over just running the tool? 

Maybe it is a powershell script and needs to be launched with powershell?

utkutombul:
Sorry for the late answer. Yes, it specifically requires PowerShell. Perhaps it would be better if `execve` situtation is explained this way; since it requires PowerShell, my first -inexperienced- initiative was to call PowerShell from `execve` but somehow it failed because of spacing. Artifact definitely does not run properly with `execve`.

scudette:
This artifact was recently updated and fixed for the latest version of cis cat.

PR #1012, opened 2025-03-07
Create Windows.Detection.Honeyfile.yaml

zaneGittins:


content/exchange/artifacts/Windows.Detection.Honeyfile.yaml
```diff
@@ -0,0 +1,203 @@
+name: Windows.Detection.Honeyfile
+author: Zane Gittins
+description: |
+   This artifact deploys honeyfiles according to the Honeyfiles CSV parameter. It first enables system-wide file auditing to monitor file access,
```
scudette: This is a really cool idea but wont it be easier to monitor with etw?

You can get file access ETW events like in this example:
https://sigma.velocidex.com/docs/models/windows_etw_base/#readfile

zaneGittins: This would be a great way to eliminate the need to setup the audit policy/acls! I tried it out using this VQL:
```
SELECT * FROM
watch_etw(guid='kernel',
   kernel_tracer_type=['file','process'],
   kernel_tracer_stacks=['file','process'])
WHERE System.KernelEventType = "ReadFile" AND EventData.FileName =~ ".kdbx"
LIMIT 5
```
But the events I got looked like this:
image

They look good but missing the process information that accessed the file, is there anything I need to do to get the process information (ProcInfo) to show up?

zaneGittins: Ah I see this example here: https://github.com/Velocidex/velociraptor/blob/4365156edcf2c13493a5e105cea3ba9551e93f83/artifacts/definitions/Windows/ETW/KernelFile.yaml#L52

I think I can do similar with the kernel logger/file tracker, I'll try this out.

zaneGittins:
@scudette This version should be working now using ETW events! Thanks for the suggestion!

mgreen27:
This is a really cool artifact. We can probably also do the file write in VQL and remove the need for powershell at all - here is an example of random character creation - https://docs.velociraptor.app/knowledge_base/tips/random/

zaneGittins:
@mgreen27 that's a great idea! Would be cool if we could do it without shelling out. Any ideas for functions to use for writing the file and the magic bytes? I see tempfile() but nothing for a permanent file.

mgreen27:
We can use the copy() function and just jam the first magic bytes in manually before the random ones. 
I can take a look tomorrow - I will need to play around, the example I gave you I used for generating a random password but we can do the same for files and expand the range.

mgreen27:
@zaneGittins I got the file creation working pretty well - deploying to a specific honeyuser and regex replacing the path. I think we just need to figure out the logic on creation.
We were thinking A check the user exists + file exists so people dont accidently overwrite existing files.
image

zaneGittins:
@mgreen27 This looks great, very cool we are able to do this in VQL! That sounds like a good approach to me, and if the file already exists, we should still monitor access events for when the client restarts/etc.

zaneGittins:
@mgreen27, also maybe we can use users() to get all users and feed them to your file creation? I think we'd want to deploy to all users on the client to cast the widest net

zaneGittins:
Perhaps something like this? Although not sure if we can rely on `FileSystems[0]` always being the primary drive or not.

```sql
Let FileSystems <= select *,format(format="%s\\Users",args=[PartitionStat.device]) as HomeRoot from filesystems()

// example honeyfile path
LET Path = "\\Documents\\KeePass\\KeePass.kdbx"

// Use size=0 to filter out files, maybe there is a better way though?
LET PathsToMonitor <= select * from foreach(row=FileSystems[0], query={select *,format(format="%s%s",args=[OSPath,Path]) as HoneyfilePath from vfs_ls(path=HomeRoot) WHERE Size = 0 AND NOT OSPath = NULL})

LET ExistsLookup <= to_dict(item={
    SELECT FullPath as _key, FullPath as _value from foreach(row=PathsToMonitor, query={select * from stat(filename=HoneyfilePath)})
})

LET PathsToCreate = SELECT HoneyfilePath, get(item=ExistsLookup,field=HoneyfilePath) as PathExists
FROM PathsToMonitor
WHERE PathExists = NULL

SELECT * from PathsToCreate
```

mgreen27:
here is my example you can test with. Takes a regex username and enables managing the files by also adding a marker so we can remove them https://gist.github.com/mgreen27/ccef8917c57a834f61292aa61347dd11

I think honeyfiles is a is a cool capability we should look at adding it into the main project. We probably need to check performance and etw multiplexing.

zaneGittins:
This is really cool, thanks Matt!! Then in my artifact it would be as simple as replacing the PowerShell with this, where default HoneyUserRegex is ".", we'd just need both these artifacts in the Exchange or main project.
```
LET WatchFiles <= SELECT *
         FROM Artifact.Windows.Detection.Honeyfiles.Manage(
                                                          DeployHoneyFiles=True,
                                                          HoneyUserRegex=HoneyUserRegex)
```
When I pass the Honeyfiles argument, I don't get any results. I did adjust the format of my Honeyfiles to match `Windows.Detection.Honeyfiles.Manage`, I'll try to pinpoint what the issue is.
```
LET WatchFiles <= SELECT *
         FROM Artifact.Windows.Detection.Honeyfiles.Manage(Honeyfiles=Honeyfiles,
                                                          DeployHoneyFiles=True,
                                                          HoneyUserRegex=HoneyUserRegex)
```

EDIT - this is very odd but if I change the header from HexString to MagicBytes in both artifacts everything works as expected, maybe some parsing issue on the word "string", but unsure. Was able to replicate the bug with a simple artifact:

```
name: Custom.Artifact.TestCSV
description: |
   This is the human readable description of the artifact.

# Can be CLIENT, CLIENT_EVENT, SERVER, SERVER_EVENT or NOTEBOOK
type: CLIENT

parameters:
  - name: TestCSV
    description: test
    type: csv
    default: |
        Test,Test2
        A,B
        C,D
        E,F

sources:
  - precondition:
      SELECT OS From info() where OS = 'windows' OR OS = 'linux' OR OS = 'darwin'

    query: |
      SELECT * from TestCSV
```

Which I call with this artifact:
```
name: Custom.Artifact.TestCallCSV
description: |
   This is the human readable description of the artifact.

# Can be CLIENT, CLIENT_EVENT, SERVER, SERVER_EVENT or NOTEBOOK
type: CLIENT

parameters:
  - name: TestCSV2
    description: The honeyfiles to generate and monitor.
    type: csv
    default: |
        TargetPath,Enabled,HexString
        A,Y,CCCCCC
        B,Y,CCCCCC

sources:
  - precondition:
      SELECT OS From info() where OS = 'windows' OR OS = 'linux' OR OS = 'darwin'

    query: |
      SELECT * from Artifact.Custom.Artifact.TestCSV(TestCSV=TestCSV2)
```

When the csv header has HexString then we don't get any rows, but if you just change it to 'Hex' we do.

zaneGittins:
I've been testing this artifact (the kernel etw variant) on around 15 Windows VMs. The kernel etw does seem to be very cpu intensive, went from around average 5% CPU to 70%+ constant on most systems.

zaneGittins:
I was able to refactor the artifact and CPU looking good again now - issue was in my artifact. Since Kernel ETW generates so many events have to be careful on how we call the process tracker as well as foreach loops, changed to a get() instead which should be O(1) instead of O(n)

scudette:
I think we should maybe include the other artifact inline as it can not be relied on that it will be available

mgreen27:
In think one of the issues of inline is we loose the ability to manage the honey files as easily? 
The other similar capability we have is deploy sysmon in the process tracker which also calls a static artifact.
We can probably join but im a bit worried if we deploy to all users and dont clean up.

scudette:
Cleanup is simple with atexit()

scudette:
The process tracker does not require sysmon - there are several types of feeds for the process tracker now (one is an etw feed)

mgreen27:
I was refering to the sysmon install component - but atexit() solves the problem.

zaneGittins:
Updated to add Matt's code inline. CPU usage is fine while idle, but it does spike when a lot of file reads are happening, not sure if there is anything we can do to improve this or not. We saw a roughly 10% average cpu increase and 80MB memory increase across the ~ 20 clients we tested against using this artifact. This is probably fine for targeted deployment, but not mass deployment across many systems.

mgreen27:
I tested and it seems to work. I cant seem to get a hit on notepad.exe though. Tested in powershell:

```
$files= ( 
        "C:\Users\yolo\Documents\KeePass\KeePass.kdbx",
        "C:\Users\yolo\AppData\Local\KeePass\KeePass.config.xml" )
foreach ( $filename in $files ) { 
    write-host "reading: $filename"
    Get-Content $filename | out-null
    notepad++ $filename
    notepad $filename
} 
```
I suspect notepad is accessing files in a different way. In anycae I think this is fine for the exchange - we should look at improving reliability if we wanted to productionise it

zaneGittins:
@mgreen27 good catch! If we switch to "CreateFile" we get events for notepad as I think this captures when notepad requests a handle to the file, but it's also a lot noisier, for example browsing to the folder with explorer generates events. Here is the changed code if you wanted to try it out:
```sql
LET AuditEvents = SELECT *
         FROM delay(
           query={
             SELECT 
                    timestamp(
                      string=System.TimeStamp) AS Timestamp,
                    get(
                      item=WatchFiles,
                      field=EventData.OpenPath) AS IsHoneyFile,
                    *
             FROM watch_etw(guid='kernel',
                            kernel_tracer_type=['file'])
             WHERE (System.KernelEventType = "CreateFile")
              AND IsHoneyFile != NULL
                   AND System.ProcessID != getpid()
           },
           delay=1)
       SELECT 
              Timestamp,
              System.ProcessID AS Pid,
              EventData.OpenPath AS FileName,
              IsHoneyFile,
              process_tracker_get(
                id=System.ProcessID).Data AS ProcInfo,
              join(
                array=process_tracker_callchain(
                  id=System.ProcessID).Data.Name,
                sep="->") AS CallChain
       FROM AuditEvents
       WHERE NOT ProcInfo.Exe =~ ProcessExceptionsRegex
 ```
 Let me know if you think this is a better route

**Comment by scudette** [comment]
Browsing to the folder with explorer actually does open and read all the files so that explorer can show the icon - unfortunately this is a common issue with many honey files is that files are sometimes opened and read for benign reasons (e.g. virus scanned etc).

I think this is an acceptable FP - maybe we can eliminate it with a filter on the call chain?

**Comment by zaneGittins** [comment]
Changed to use CreateFile, and I added Explorer.exe to the ProcessExceptionsRegex. CPU usage is still a little higher than I'd like, but I think it should capture anything reading the honeyfile now.

**Comment by mgreen27** [comment]
I tested and this works - there is a lot of CreateFile hits - im happpy using that but there are lots and lots of entries. I guess this is ok when we are targeting very specific files and we would expect the amount of activity to be low.

**Comment by zaneGittins** [comment]
It does generate a lot of entries compared to ReadFile or using event id 4663s, is there a way we can group by similar events in a CLIENT_EVENT type artifact? If possible, we might want to group by the Honeyfile and PID for a time window of a few seconds, and then add on a count() which would cut down on the total number of events being sent to the server

**Comment by zaneGittins** [comment]
I've been testing a version using Microsoft-Windows-Kernel-File instead of NT Kernel Logger, performance is WAY better and it captures notepad too. Going to do some more testing but I'll commit an updated version in a bit.
image

**Comment by scudette** [comment]
The problem is that the kernel file gives the filename as device filenames and in the NT Kernel logger we translate those back to filesystem names.

Perhaps we should also do this for the Kernel-File logger

**Comment by zaneGittins** [comment]
@scudette I added in some VQL to do the translation in that latest commit, based on what you had previously shared in the discord using winobj()

**Comment by mgreen27-r7** [comment]
your outrunning my ability to help in my free time :)

Materialization your getpid()
I also added another to AuditEvents to allow for simple filtering before you did your lookup

We might also want to use the dedup() plugin to minimize events - I'll need to mess with that.

```
LET CurrentPid <= getpid()
LET TargetEvents = SELECT *
         FROM watch_etw(guid='{edd08927-9cc4-4e65-b970-c2560fb5c289}',
                        description="Microsoft-Windows-Kernel-File",
                        any=Keyword)
         WHERE System.ID = 12
               AND System.ProcessID != CurrentPid
               
 LET AuditEvents = SELECT 
                                timestamp(
                                  string=System.TimeStamp) AS Timestamp,
                                get(
                                  item=WatchFiles,
                                  field=EventData.FileName) AS IsHoneyFile,
                                *
         FROM TargetEvents
         WHERE IsHoneyFile != NULL
```


Also add in a WHERE clause for IsHoneyFile as in some cases we would not want to monitor the honeyfile if it is a legitimate application accessed legitimately (unless you do).

```
LET WatchFiles <= to_dict(
           item={
             SELECT 
                    KernelPath AS _key,
                    IsHoneyFile AS _value
             FROM foreach(
               row=KernelVolumes,
               query={
                 SELECT *,
                        regex_replace(
                          source=TargetPath,
                          replace=SymlinkTarget,
                          re="[A-Z]+\:") AS KernelPath
                 FROM add_honeyfiles
                 WHERE TargetPath =~ UserDrive
                  AND IsHoneyFile
               })
           })
```

**Comment by scudette** [comment]
I am actually interested to know why the CPU is more used in the NT Kernel logger case than the KernelFile case. They should be fairly similar.

Can you see how many events were generated from each source? Maybe a CPU profile will be good?
You can see what the client is doing with ETW by collecting the following

```
SELECT * FROM profile(type="ETW")
```

zaneGittins:
I ran both versions (NT Kernel Logger and Kernel File) concurrently, and then took an ETW profile, the NT Kernel Logger does seem to generate many more events (17167 vs 517966):
image
I ran the test for a duration of 5 minutes, and browsed around in explorer, as well as commands in powershell to generate activity. I also generated CPU profiles individually for each artifact and shared in a DM in Discord.

zaneGittins:
Thanks @mgreen27! I added in those changes, I also removed the winobj() as with the latest Velociraptor build this is no longer necessary, thanks Mike! Also moved the atexit() below remove_honeyfiles, for it to work properly. KernelFile is also less noisy, so we might be able to get away without dedup, but let me know what you think.

PR #1014, opened 2025-03-12
Create Linux.Detection.Honeyfiles.yaml

zaneGittins:
Variant of honeyfiles for Linux using eBPF.


zaneGittins:
We may want to set permissions of the files to readable by all 644, but unsure how to do this without shelling out. The copy() permissions argument seems to only support 600, 700 or 400. 
```go
...
permissions := os.FileMode(0600)

	switch arg.Permissions {
	case "x":
		permissions = 0700

		// On windows executable means it has a .exe extension.
		if runtime.GOOS == "windows" &&
			!strings.HasSuffix(arg.Destination, ".exe") {
			arg.Destination += ".exe"
		}

	case "r":
		permissions = 0400
	}
...
```
EDIT: Or at the very least we may want to chown the file to the correct user.

PR #1025, opened 2025-04-04
Add artifact listing VS Code extensions

misje:
This is a rather simple artifact, but it is still useful. The information gathered would be more useful if tied together with extension information fetched from the unofficial marketplace API. I have a nearly finished notebook suggestion that does just that, but due to the sheer number of extensions (+72k) and API rate limits, I am not sure if it is stable or useful. Example:

### DIFF
--- a/content/exchange/artifacts/Generic.VSCode.Extensions.yaml
+++ b/content/exchange/artifacts/Generic.VSCode.Extensions.yaml
@@ -1,0 +1,84 @@
+name: Generic.VSCode.Extensions
+author: "Andreas Misje – @misje"
+description: |
+  List all installed Visual Studio code extensions
+
+  This artifact lists all installed VS Code exensions on all supported
+  operating systems. This list can be used to hunt for known malicious
+  extensions by name or ID.
+
+  The artifact returns the raw JSON from the extensions file. The notebook
+  provides a more user-friendly view.
+
+type: CLIENT
+
+parameters:
+  - name: ExtensionGlobs
+    description: |
+      Where to look for extension.json files on varios OSes
+    type: csv
+    default: |
+      OS,Glob
+      windows,C:\Users\*\.vscode\extensions\extensions.json
+      linux,/home/*/.vscode/extensions/extensions.json
+      darwin,/Users/*/.vscode/extensions/extensions.json
+
+column_types:
+  - name: ID
+    description: Extension name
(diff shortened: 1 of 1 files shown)


misje:
Here is an example of the complete list of extensions fetched from the unofficial marketplace API, which probably has many use cases. Even if its metadata is not used at all, it could be used to cross-reference extensions that has been removed/revoked from the marketplace(?):

scudette:
If we are restricted by API limitations we can mirror the API in a JSON file and host it statically on github. This would make it a useful resource and allow everyone to use it easily

content/exchange/artifacts/Generic.VSCode.Extensions.yaml
```diff
@@ -0,0 +1,84 @@
...
+      SELECT *
+      FROM foreach(
+        row={
+          SELECT OSPath
+          FROM glob(globs=Globs.Glob)
+        },
+        query={
+          SELECT OSPath,
+                 parse_json_array(data=Data) AS Extensions
+          FROM read_file(filenames=OSPath)
```
scudette: Would it be better to just upload the files rather than include them in the JSON blob? We try to avoid that as it can explode the size of the JSON responses and lead to issues in the GUI.

You can use preview_upload column type to preview the file in the GUI as well which makes it a lot nicer

misje:
> If we are restricted by API limitations we can mirror the API in a JSON file and host it statically on github. This would make it a useful resource and allow everyone to use it easily

That is also an alternative. I will look into that in the next pull request adding this feature.

> Would it be better to just upload the files rather than include them in the JSON blob? We try to avoid that as it can explode the size of the JSON responses and lead to issues in the GUI.

Personally I do not think so. The files are typically very small. In my recent hunt, the results show that most have a few extensions installed, some about 20. The most extreme case is 78. Each extension entry is not large. Does this warrant file uploads instead? I have yet to experience GUI issues due to large JSON data in any of the artifacts I have ever used, but I would like to know more about when it becomes an issue, and adapt artifacts if needed.

scudette:
Yes it is always necessary to upload the files - then you get the preview box for free and can export them easily as files. 

Including bulk data inside JSON is almost always a bad idea

misje:
I have seen interest in an artifact like this several times in Discord since it was written, so I am re-opening this. Modifications are welcome.

PR #1033, opened 2025-04-16
changed hugo docker image to official

SolitudePy:
changed hugo docker image to official


SolitudePy:
*this is actually a community project but super up to date

PR #1079, opened 2025-07-10
Created Windows.Registry.ProcessMitigationOptions.yaml

bmcdermott-r7:
Added an artifact that parses the security mitigation options within the Windows Registry.

### DIFF
--- a/content/exchange/artifacts/Windows.Registry.ProcessMitigationOptions.yaml
+++ b/content/exchange/artifacts/Windows.Registry.ProcessMitigationOptions.yaml
@@ -1,0 +1,21 @@
+name: Windows.Registry.ProcessMitigationOptions
+description: |
+   This artifact parses Exploit Guards MitigationOptions to see what security controls
+   are enabled for each process, and returns them in a 15 digit hexidecimal value. 
+   
+   Ref: https://theryuu.github.io/ifeo-mitigationoptions.txt
+
+type: CLIENT
+
+sources:
+  - precondition:
+      SELECT OS From info() where OS = 'windows'
+
+    query: |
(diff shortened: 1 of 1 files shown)


content/exchange/artifacts/Windows.Registry.ProcessMitigationOptions.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+type: CLIENT
+
+sources:
+  - precondition:
+      SELECT OS From info() where OS = 'windows'
+
+    query: |
+        SELECT 
+            path_split(path=OSPath)[-2] AS ProcName, OSPath,
```
scudette: You don't need path split here just ospath[-2] is fine

scudette (commented):
We generally want to add these to the registry hunter instead as that includes a lot of rules already. I can add it to it now if you like instead

bmcdermott-r7:
@scudette Happy for it to be added to RegistryHunter, but I think there's still a place for smaller precise artifacts. I sent you a DM, but we have difficulties with these larger artifacts potentially because of the scale of our network.

scudette:
Thanks for the PR - the rule was added here 
https://github.com/Velocidex/registry_hunter/blob/a41a9af3392a8d97f88d9dd5887c550065a5ddab/Rules/Velociraptor-Rules.yaml#L585

I also added a full parse of the bits according to the reference you provided.

If you just want to collect that one rule, simply restrict the registry hunter DescriptionFilter to "MitigationOptions"

You can see the rules and search for which one you need using the rules site https://registry-hunter.velocidex.com/docs/rules/

This is an example of the output
image

PR #1129, opened 2025-10-21
Make Linux honeyfiles readable

zaneGittins:
Linux honeyfiles were originally only readable by root (from the "copy" command), this change uses chmod to make them readable by everyone.

### DIFF
--- a/content/exchange/artifacts/Linux.Detection.Honeyfiles.yaml
+++ b/content/exchange/artifacts/Linux.Detection.Honeyfiles.yaml
@@ -33,121 +33,135 @@ sources:
         SELECT OS From info() where OS = 'linux' AND version(plugin="watch_ebpf") >= 0
 
     query: |
-       LET RandomChars(size) = SELECT
-           format(format="%02x", args=rand(range=256)) AS HexByte
-         FROM range(end=size)
-       
-       LET check_exist(path) = SELECT
-           OSPath,
-           Size,
-           IsDir,
-           if(condition=read_file(filename=OSPath)[-7:] =~ 'VRHoney',
-              then=True,
-              else=False) AS IsHoneyFile
-         FROM stat(filename=path)
-       
-       LET enumerate_path = SELECT regex_replace(source=TargetPath,
-                                                 re='''\%USERPROFILE\%''',
-                                                 replace=Homedir) AS TargetPath,
-                                   *,
-                                   check_exist(path=regex_replace(
-                                                 source=TargetPath,
-                                                 re='''\%USERPROFILE\%''',
-                                                 replace=Homedir))[0] AS Exists,
-                                   MaxSize - rand(range=(MaxSize - MinSize)) -
-                                     len(list=unhex(string=MagicBytes)) - 7 AS _PaddingSize
-         FROM Honeyfiles
-       
-       LET target_users = SELECT User,
-                                 Homedir,
-                                 Uid
-         FROM Artifact.Linux.Sys.Users()
-         WHERE int(int=Uid) >= 1000
-          AND NOT Homedir = '/nonexistent'
-               AND User =~ HoneyUserRegex
-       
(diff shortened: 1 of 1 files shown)


zaneGittins:
@scudette and @mgreen27, I've noticed an issue with this artifact. When we create directories for honeyfiles on Linux using the copy function with create_directories='y', the directories are set to `drwx------` allowing no access other than by the owner (root). Honeyfiles on Linux work great where the directory already exists, however when Velociraptor creates the directories using copy the honeyfiles aren't readable by most users, and unlikely to be triggered even by a threat actor.  I think ideally we want the directories to be 755 and the files to be 644, owned by root is fine. I've been struggling to find a way to set only directories created by the artifact to 755 while not affecting preexisting directories. For example if we have `%USERPROFILE%/.aws/credentials` only .aws folder should be set to 755, not the users home folder. I wanted to check if either of you have any ideas on how we might address this? I've been trying to find an elegant solution, but really struggling on the best way to handle this.

scudette:
This is a good observation! The artifact relies on the create_directories parameter to copy() but that was originally used to e.g. unpack temporary files for our own use so we always create directories with very locked down permissions.

This use case is a bit different so maybe we need to carry over the file permissions to the directory permission parameter?

scudette:
Actually looking at the permissions again we don't support the file being readable at all:

default file permission is 0600 or 0700 and always 0700 for directories.

zaneGittins:
> Actually looking at the permissions again we don't support the file being readable at all:
> 
> default file permission is 0600 or 0700 and always 0700 for directories.

I noticed that too and tried to fix by shelling out to call chmod to make the files readable, but I'm struggling when the artifact makes directories, its tricky for me to differentiate between folders that were preexisting, and those made by the copy function, if it's possible to add additional permission options to the copy() function, then I think that would make this much easier

scudette:
Yes absolutely - this is probably a feature creep from the original use case of copy() which was used to just extract something from an external tool or something like that. 

The question is if we should also add a chmod() function or split it into a mkdir() function + copy() or just add another parameter to copy()

zaneGittins:
At least for the Honeyfiles, I think adding another parameter to copy() as well as adding a chmod() function would work well. Then we can set the permissions in copy() to be 755, and use chmod() afterwards to set the file to 644, which would avoid shelling out.

zaneGittins:
Another issue I've noticed with this artifact, on some systems (for example web servers) using security_file_open can cause very high cpu. I think ideally, we'd want to filter at the kernel level for these honeyfile paths, but that would require implementing tracee filters.  Using security_file_open without filters works for now, just need to be careful which systems it's used on.

I think this is also sometimes the source of high cpu when using artifact Linux.EBPF.Monitoring, since Linux.Sigma.EBPFBase collects security_file_open.

scudette:
Do you have evidence that implementing the filtering in the ebpf layer improves overall performance much? You can try to use tracee itself on those systems (I think it allows for filtering but I am not sure if it does it in kernel or user space).

I am not so sure that ebpf can perform the filtering in kernel space any more efficiently, but maybe there is overheads due to VQL filtering as opposed to raw golang filtering.

Do you know what is the event rate? (you can see it in the profile server's ebpf section).

zaneGittins:
I profiled for 30 seconds, and it shows 2107481 total events (although not all would be security_file_open).

image

I've reviewed the Tracee repo and docs, and I'm pretty sure the filtering does happen in userspace even in tracee.. but tracee does seem more performant then velociraptor for security_file_open

```bash
./tracee-ebpf-static -e security_file_open.data.pathname=/home/* -o out-file:/tmp/test2
```
vs.
```
SELECT * FROM watch_ebpf(events=["security_file_open"])
WHERE EventData.pathname =~ "/home/"
```

Velociraptor jumps from ~5% to 20% when running the above VQL, where running tracee uses negligible cpu in my test. 

EDIT: Going to do some more testing across a few other systems to verify this is accurate.

PR #1132, opened 2025-10-27
Add Windows.Forensics.QuickAssist artifact

BoriBori:
This artifact investigates Quick Assist activity by searching prefetch files, parsing identified files, checking EdgeWebView history, and searching event logs for related activity.


BoriBori:
I updated the artifact to use the built-in prefetch artifact and the built-in sqlite db parser

PR #1154, opened 2025-12-12
Add Linux.Events.EBPFEnriched artifact

zaneGittins:
This artifact captures and enriches EBPF events on Linux endpoints with process, Docker container, and user details.

### DIFF
--- a/content/exchange/artifacts/Linux.Events.EBPFEnriched.yaml
+++ b/content/exchange/artifacts/Linux.Events.EBPFEnriched.yaml
@@ -1,5 +1,4 @@
 name: Linux.Events.EBPFEnriched
-name: Custom.Linux.Events.EBPF
 author: Zane Gittins
 description: |
  This artifact forwards EBPF events generated on the endpoint and enriches events with:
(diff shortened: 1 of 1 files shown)


zaneGittins:
This artifact will need an update to also use HostProcessID and HostParentProcessID if this is fixed in the process tracker https://github.com/Velocidex/velociraptor/issues/4596

zaneGittins:
Updated artifact to use HostProcessID and HostParentProcessID, added further enrichment of usernames from docker containers and hashing on execution within docker containers using overlayfs.

zaneGittins:
I didn't realize that System.Hostname is a max of 15 chars from code: (`Set("HostName", string(bytes.TrimRight(eCtx.UtsName[:], "\x00"))).` 

Made the following changes to the artifact:

- Fixed username enrichment for hosts with hostname > 15chars.
- Made BufferSize for delay() a parameter.
- BufferSize for delay() increased from 2500 to 3000
- Excluded magic_writes to /sys/fs/cgroup/ which are very bursty and blow our delay() buffer
- Moved all exclusions to exclusions parameter.

zaneGittins:
Found that the overlayfs enrichments only work on docker v29+, I don't think it's an issue, but I should note it in the artifact.

zaneGittins:
I think this is my final update for now, but this may need an update if events is changed on watch_ebpf()

PR #1158, opened 2025-12-15
Update syncAsset query to use Exchange namespace

hasamba:


scudette:
We don't actually use the exchange namespace any more. 

By default all artifacts are imported without a namespace

PR #1160, opened 2025-12-22
Add Windows DFIR script for data collection

go-LANz:
Adds a Velociraptor artifact that runs @Bert-JanP DFIR script https://github.com/Bert-JanP/Incident-Response-Powershell
And upload the results back to the server


scudette:
There are  many one off type IR scripts around . We obviously dont really want to host hundreds of such artifacts on the exchange. Generally we prefer to have these type of triage scripts built into VQL (running powershell needlessly tends to clobber many forensic artifacts).

Although we do have artifacts the run external triage tools (like CyberTriage, UAC etc) I wonder if we should reduce these to only launch the main and most complete ones. Is this script more complete/better than others? Would it be of interest to most people who download the exchange?

Of course there is nothing stopping the author of this script from publishing this artifact on their release pages - it is just that it might not be of interest to the main audience of the exchange.

PR #1165, opened 2026-01-03
Create keyboard inputlocales artifact

Guzzy711:


mgreen27:
I think you definitely want to have a MTime output for each row on this one.
I cant comment too much on workflow at scale yet for this one, I have found also outputting some of the other fields in the system keys are helpful context.

scudette:
* This is probably better written with read_reg_key() 
* I think this needs to go in the registry hunter rather than its own separate artifact. Probably we dont want to have thousands of separate artifacts for registry artifacts.

Guzzy711:
Sure, I will look at making it use the registry hunter. But from reading the docs, is it required to import the registry hunter, or is it part of newer versions of Velociraptor?

scudette:
I added the rules to the registry hunter here https://github.com/Velocidex/registry_hunter/blob/a7e0abc27274d84bde40f4dbbbb5eb4faa71faf2/Rules/Velociraptor-Rules.yaml#L912

i also realized that there is no document that helps people when they want to add new rules - so I made it here
https://registry-hunter.velocidex.com/docs/registry_hunter/develop/

It is better to use the registry hunter for this artifact because it takes care of the user hives by itself - this artifact does not handle the case where a user is not currently logged in and the raw registry hive needs to be parsed.

Guzzy711:
I finally had the time to look at the registry hunter. I like the idea about the whole idea about rules and making it easier to write artifacts without reinventing the wheel. Will def use that for the future. In terms of this PR, what would be the approach, since I guess it doesn't really make sense to merge this artifact into the exchange, hence we got it in the registry hunter library?

scudette:
We have a to-do to remove pure registry artifacts from the exchange in favor of the registry hunter - I think we should drop this PR.

Guzzy711:
Aight, let's go with that. 😄

PR #1166, opened 2026-01-04
Add RdpClientInfo artifact for RDP session analysis

Guzzy711:
This artifact extracts client-side characteristics recorded during RDP session connections, including timezone configuration and operating system type information.

### DIFF
--- a/content/exchange/artifacts/RdpClientInfo.yaml
+++ b/content/exchange/artifacts/RdpClientInfo.yaml
@@ -1,0 +1,70 @@
+name: Windows.EventLogs.RdpClientInfo
+author: Guzzy (blog.guzzy.dk) | SagaLabs
+
+description: |
+  The artifact extracts client-side characteristics recorded during RDP session connection, including timezone configuration (Event ID 104) and operating system type information (Event ID 169) from the source system.
+  The artifact is based on research by @thedigitaldetective, and article is referenced below.
+
+  The artifact uses EvtxHunter artifact to query Microsoft-Windows-RemoteDesktopServices-RdpCoreTS/Operational.evtx.
+
+  Event ID 104 records the client timezone bias, which is a UTC offset and used to infer likely geographic regions
+  of the RDP client. For example if it says [1] it means the client is set to UTC+1 timezone.
+
+  Event ID 169 records client OS details, including osMajorType and osMinorType values, which can assist in identifying the RDP client platform.
+  To look up the OS type values, refer to the Microsoft RDP documentation linked below.
+
+  MajorType values:
+
+  - 0: Unspecified platform
+  - 1: Windows platform
+  - 2: OS/2 platform
+  - 3: Macintosh platform
+  - 4: UNIX platform
+  - 5: iOS platform
+  - 6: OS X (macOS) platform
+  - 7: Android platform
+  - 8: Chrome OS platform
+
+  For example if MajorType=1 and MinorType=3, this indicates that the RDP client is a Windows NT–based operating system.
+
+reference:
+  - https://medium.com/@thedigitaldetective/rdp-forensics-part-2-fingerprinting-attacks-with-timezone-and-monitor-display-resolution-3ebe668d52ad
+  - https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/41dc6845-07dc-4af6-bc14-d8281acd4877
+
+type: CLIENT
+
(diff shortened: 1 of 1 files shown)


mgreen27:
I have an internal artifact for this but also include the other EIDs and features like time boxing. 
Its useful when tracking actors in specific live cases. In practice, grabbing the extra eids in one view is better as you can manipulate data to align with sessions.

LMK if you are ok for me to add to this.

Guzzy711:
@mgreen27 sounds cool! 
I'm totally cool with you adding this to the artifact - excited to see it. :-)

mgreen27:
cool - there is not much to this one, mostly in the analysis.

Run once once to collect historical then group by and find outliers.
I work in a interesting customer base timezone where we often see the attacker timezone's pop out, I have started tracking this somewhat.


```
name: Windows.EventLogs.RDPIntel
author: "Matt Green - @mgreen27"
description: |
    Grab RDP intel artifacts from Microsoft-Windows-RemoteDesktopServices-RdpCoreTS/Operational.evtx
    
    - EID 104 records the UTC timezone offset of the client system initiating the connection.
    - EID 131 records IP address - I have added this to enable discerning sessions .
    - EID 168 records resolution of desktop width and height information
    - EID 169 records both the operating system platform type (osMajorType) as 
    well as the platform version type (osMinorType) of the client system 
    initiating the connection.
    
    Combine with RDPAuth or standard EvtxHunter to scope RDP and authentication events.

reference:
    - https://medium.com/@thedigitaldetective/remote-desktop-protocol-using-client-keyboard-input-in-attack-attribution-and-profiling-94a76f0f4ff4
    - https://medium.com/@thedigitaldetective/rdp-forensics-part-2-fingerprinting-attacks-with-timezone-and-monitor-display-resolution-3ebe668d52ad


precondition: SELECT OS From info() where OS = 'windows'

parameters:
  - name: TargetGlob
    description: path to event log.
    default: '%SystemRoot%\System32\Winevt\Logs\*RDPCoreTS*.evtx'
  - name: TargetIDRegex
    description: EventID to target
    default: ^(104|131|168|169)$
  - name: VSSAnalysisAge
    type: int
    default: 0
    description: |
      If larger than zero we analyze VSS within this many days
      ago. (e.g 7 will analyze all VSS within the last week).  Note
      that when using VSS analysis we have to use the ntfs accessor
      for everything which will be much slower.
  - name: DateAfter
    type: timestamp
    description: "search for events after this date. YYYY-MM-DDTmm:hh:ssZ"
  - name: DateBefore
    type: timestamp
    description: "search for events before this date. YYYY-MM-DDTmm:hh:ssZ"

sources:
  - query: |
        SELECT * FROM Artifact.Windows.EventLogs.EvtxHunter(
            EvtxGlob=TargetGlob,
            IdRegex=TargetIDRegex,
            DateBefore=DateBefore,
            DateAfter=DateAfter,
            VSSAnalysisAge=VSSAnalysisAge )

    notebook:
      - type: vql_suggestion
        name: Timezone by IP
        template: |
            /*
            ### Timezone bias and ClientIP data
            */
            SELECT EventTime,  Computer, Provider,EventID,
                EventData.ClientIP as ClientIP,
                EventData.ConnType as ConnType,
                EventData.Data.Value as TimeZoneBias
            FROM source()
            WHERE EventID in ( 104, 131 )
      - type: vql_suggestion
        name: Groupby counts
        template: |
            /*
            ### Group RDP intel data by EventData
            */
            SELECT 
                min(item=EventTime) as FirstSeen,
                max(item=EventTime) as LastSeen,
                Computer, EventID,
                EventData,
                count() as Total
            FROM source()
            GROUP BY EventData
```

Guzzy711:
Hi @mgreen27 - cool artifact. 
Do you think this should live as its own artifact, or would you prefer I adapt mine to align with your approach? I’m good with whichever you think makes the most sense from a usability and analysis standpoint.

scudette:
What is the difference between the two? IMHO It would be better if we consolidated them into one

Guzzy711:
I think Matts is more detailed and "battle-tested" than mine. 
However I personally like that the artifact is split into different sources for the different event ids, e.g. one table for timezone, one for Osinfo etc...

What is your take on this?
image
image

mgreen27:
Its collect once and notebook or collect per usecase. 
What happens when you use this artifact is you want to pull some of the other EIDs in to do analysis with regularly to see the other config options.

PR #1168, opened 2026-01-08
Create Windows.Sys.StartupItemsWithHash.yaml

Yoon1029:
Applications that will be started up from the various run key locations. This artifact extracts file hashes for all referenced executables.
    
Modified from Windows.Sys.StartupItems to include hash values for registry-based startup entries.

### DIFF
--- a/content/exchange/artifacts/Windows.Sys.StartupItemsWithHash.yaml
+++ b/content/exchange/artifacts/Windows.Sys.StartupItemsWithHash.yaml
@@ -1,0 +1,214 @@
+name: Windows.Sys.StartupItemsWithHash
+description: |
+    Applications that will be started up from the various run key
+    locations. This artifact extracts file hashes for all referenced executables.
+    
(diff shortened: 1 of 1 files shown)


scudette:
I believe these are all covered pretty well by the registry hunter:

https://registry-hunter.velocidex.com/docs/rules/#Run%20(NTUSER)
https://registry-hunter.velocidex.com/docs/rules/#Run%20(SYSTEM)

And I think the Lnk one is covered by the triage artifact
https://triage.velocidex.com/docs/windows.triage.targets/rules/#LnkTargets

If you find any gaps we can add to those artifacts.

scudette:
Does this give more functionality than the registry hunter or triage artifacts? Those have much more robust adaptive target detection and can hash as well as download the targets of the run keys.

scudette:
Does this add to the regular registry hunter? That has a lot of rules but you can just look at run keys by having a rule filter like "Run"

image


The registry hunter handles things like ntuser hives correctly and also checks for signing and other things.

PR #1177, opened 2026-01-21
Optional hostname addition to hayabusa result file.

julianghill:
Adds an option to include the hostname in the output CSV/JSONL filename. The default remains False to avoid impacting existing workflows.

This is useful for automated ingestion (e.g., OpenRelik) where the hostname can be used downstream (like Timesketch sketch naming via API) without extra pre‑processing.

Please let me know if you’d like any changes.

### DIFF
--- a/content/exchange/artifacts/Windows.EventLogs.Hayabusa.yaml
+++ b/content/exchange/artifacts/Windows.EventLogs.Hayabusa.yaml
@@ -85,10 +85,9 @@ parameters:
    description: "Scan only common Event IDs for quicker scans"
    type: bool
    default: N
- - name: AddHostnameToFilename
-   description: "If checked, prepends the Hostname to the uploaded file."
-   type: bool
-   default: N
+ - name: OutputFilenameTemplate
+   description: "Upload filename template. Supports env vars via expand() and {ext} for csv/jsonl (ex: %COMPUTERNAME%_hayabusa_results.{ext})."
+   default: hayabusa_results.{ext}
  - name: TimeOffset
    description: "Scan recent events based on an offset (ex: 1y, 3M, 30d, 24h, 30m)"
  - name: TimelineStart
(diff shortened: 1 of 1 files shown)


scudette:
Would it make more sense to just collect generic client info in another artifact source? That way the info is there and can be confered separately if needed but it doesn't affect anything else.

julianghill:
> Would it make more sense to just collect generic client info in another artifact source? That way the info is there and can be confered separately if needed but it doesn't affect anything else.

I agree it makes sense to keep the Hayabusa artifact focused on its core job and avoid coupling it to naming.

That said, for large hunts the output filename becomes a practical issue when ingesting into tools like Timesketch/OpenRelik: if multiple endpoints produce the same hayabusa_results.csv name, uploads can collide/overwrite depending on the ingestion workflow. Today, being able to adjust the output name at collection time avoids needing a preprocessing step.

Without this, the workflow becomes:

- extract all hayabusa_results.csv files from the hunt results,
- rename them (e.g., based on client/hostname or the directory they came from),
- then upload to Timesketch/OpenRelik/etc.

Whereas with a configurable output name, you can extract and ingest directly.

If the preference is to keep this artifact “clean,” I’m also fine with leaving it as is and implementing this behavior in a separate/custom artifact. I’m mentioning it because teams using OpenRelik (e.g., LimaCharlie) have specifically asked for configurable Hayabusa result filenames, and having the same capability in a Velociraptor artifact would be useful for those pipelines.

Happy to adjust with whatever approach you think is most appropriate here, core artifact vs. separate/custom artifact.

scudette:
This is fine and makes sense. 

Do you think a better solution might be to have a filename template which is passed through expand() to determine the final name? 

Then it can use the host name or machine name from the env

julianghill:
> This is fine and makes sense.
> 
> Do you think a better solution might be to have a filename template which is passed through expand() to determine the final name?
> 
> Then it can use the host name or machine name from the env

Thanks for the good feedback.

I just made the change and hope this matches what you were thinking of.

PR #1205, opened 2026-03-15
Added page for local docs search

predictiple:
Also:
- added draft page for 0.76 release notes.
- added sections explaining free-form args and argument unpacking.
- revised advice for export timeouts to use new `Server.Utils.ExportHunt` artifact, and discourage changing the server default timeout.
- updated some info and screenshots about working with tables and GUI customization.


content/docs/vql/fundamentals/_index.md
```diff
@@ -79,6 +79,52 @@ type, and if they are required or optional.
...
+
+```vql
+LET args <= dict(root="c:/windows", globs="*")
+
+SELECT *
+FROM glob(`**`=args)
+```
+
+Note that the `**` has to be enclosed in backticks because
+[it is an identifier that uses non-alphanumeric characters]({{< ref "/docs/vql/fundamentals/#identifiers-with-spaces" >}}).
```
misje, line 101: Also note that all other arguments are ignored when `**` is used.

predictiple: Thanks I'll add that!

PR #1221, opened 2026-03-28
Add support in HoneyFiles for systems with (mostly) remote users

misje:
This artifact is really cool, but it currently only works for local users with home directories fetched from `/etc/passwd`. A new optional parameter now simply iterates home directories instead. Now, honey files can be created for remote users that has previously logged onto the system as well.

This solution works for systems with sssd and mkhomedir, and probably similar setups as well. A single glob is deemed sufficient for most use cases, e.g. "/home/*". The regex HoneyUserRegex is still applied.


misje:
I am considering dropping this pull request in favour of a rewrite of the artifact. I want to stick with just globbing the home dir, adding an epbf policy limiting events to this dir (since there are currently a lot of events), adding correct ownership and permissions to the created directories and files (since they are currently inaccessible by the current user) and a few other things.

PR #1227, opened 2026-04-01
Use latest release in spellchecker

misje:
Otherwise artifacts using default VQL arguments fail this action.


scudette:
This test failed because Windows.AttackSimulation.AtomicRedTeam was larger than 100k - we have a hard coded limit in 0.76 on the size of the artifact we will verify. 

I think it is a bit unreasonable to have such a large artifact (the data portion should be compressed to make it smaller) so I removed it for now. If anyone needs this artifact please LMK so I can fix it.

PR #1228, opened 2026-04-01
Update Linux.Detection.Honeyfiles.yaml

zaneGittins:
This update to honeyfiles uses the new Tracee policy for much better performance, as well as the new copy() permissions and dir_permissions, making honeyfiles on Linux fully functional.

### DIFF
--- a/content/exchange/artifacts/Linux.Detection.Honeyfiles.yaml
+++ b/content/exchange/artifacts/Linux.Detection.Honeyfiles.yaml
@@ -45,6 +45,8 @@ sources:
         SELECT OS From info() where OS = 'linux' AND version(plugin="watch_ebpf") >= 2
 
     query: |
+      LET ComputerName <= dict(H={ SELECT Hostname FROM info() }).H[0].Hostname
+      
       // Get detailed process information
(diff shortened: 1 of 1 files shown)


misje:
Since /home/* is now used in the profile, what about getting rid of Linux.Sys.Users() and glob home directories (like in #1221)? Then this artifact also works for users not in passwd.

zaneGittins:
Great idea! Going to make a few minor additional tweaks and I'll post an updated version

zaneGittins:
Switched to glob on home directories, and added better process information

zaneGittins:
Added Username and ComputerName fields. The process tracker lookup seems to be working unreliably, at least on the single system I tested on (v0.76.1)

misje:
> Added Username and ComputerName fields. The process tracker lookup seems to be working unreliably, at least on the single system I tested on (v0.76.1)

This might just be the issue resolved [here](https://github.com/Velocidex/velociraptor/pull/4766): Both this event artifact and the process tracker would register a policy with identical names if they start simultaneously. See if a recent build helps.

zaneGittins:
@misje yup, thanks! That fixed it

PR #1238, opened 2026-04-28
Add OneNote Hunt artifact for Windows Forensics

SwiftSecur:
This YAML file defines a hunt for OneNote files, including parameters for searching and detecting embedded images and keywords.


content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -0,0 +1,279 @@
...
+      Mtime,
+      Btime,
+      Ctime,
+      Atime,
+      hash(path=OSPath).SHA256 AS SHA256,
+
+      read_file(
+        filename=OSPath,
+        length=MaxRegexReadBytes
+      ) =~ SearchRegexAscii AS RegexAsciiMatched,
```
scudette: you can combine those regex together

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -0,0 +1,279 @@
...
+
+      read_file(
+        filename=OSPath,
+        length=MaxRegexReadBytes
+      ) =~ SearchRegexUtf16 AS RegexUtf16Matched
+
+    FROM onenote_files
```
scudette, line 81: Add the WHERE clause here to avoid hash() of files that dont match

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -0,0 +1,279 @@
...
+        condition:
+          $onenote at 0 or any of ($png,$jpg,$gif1,$gif2,$bmp,$tif1,$tif2,$webp)
+      }
+    '''
+
+    LET onenote_files = SELECT OSPath, Name, Size, Mtime, Btime, Ctime, Atime
+    FROM glob(globs=TargetGlob)
+    WHERE NOT IsDir
+      AND Size <= MaxFileSize
+      AND format(
```
scudette: Precalculate the binary string to avoid calling format() on each file.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -0,0 +1,279 @@
...
+
+      read_file(
+        filename=OSPath,
+        length=MaxRegexReadBytes
+      ) =~ SearchRegexUtf16 AS RegexUtf16Matched
+
+    FROM onenote_files
+
+
+    LET grouped <= SELECT
```
scudette: there is probably no point in the group by - it is almost guaranteed to have a single row per group because the group is so detailed. You generally try to avoid group by as much as possible as this is unlikely to be performant.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -0,0 +1,279 @@
...
+      Ctime,
+      Atime,
+      SHA256,
+      RegexAsciiMatched,
+      RegexUtf16Matched
+
+
+    SELECT
+      *,
+      if(
```
scudette: you can write it simply as UploadOriginal && upload(file=OSPath) AS OneNoteUpload

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -0,0 +1,279 @@
...
+        condition:
+          any of them
+      }
+    '''
+
+    LET onenote_files = SELECT OSPath, Name, Size
+    FROM glob(globs=TargetGlob)
+    WHERE NOT IsDir
+      AND Size <= MaxFileSize
+      AND format(
```
scudette: pre-calc here too

scudette (commented):
This is probably not much more complex than the built in file finder - consider just calling it with preset yara rules. It will be much simpler and more maintainable (and also faster).

SwiftSecur:
> This is probably not much more complex than the built in file finder - consider just calling it with preset yara rules. It will be much simpler and more maintainable (and also faster).

Thanks for all the input, really helpful, learnt a few things. I will update and see where we get to.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
SwiftSecur: Thanks, I will define this further up and reference it to avoid calling format on everything.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
SwiftSecur: thanks - thats insightful and helpful.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
SwiftSecur: added a fix for this.

SwiftSecur:
Added in an update that hopefully has some improvements on the last. Thanks for the feedback and help identifying performance issues.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
scudette: What I mean by pre-calculate is that the OneNoteMagic is a constant so you dont have to call format on each row to compare to it - just create the binary string in advance then compare directly

```sql
LET OneNoteMagic <= unhex(string='e4525c7b8cd8a74daeb15378d02996d3')


....
WHERE NOT IsDir
      AND Size <= MaxFileSize
      AND read_file(filename=OSPath, length=16)=OneNoteMagic
```

SwiftSecur: Ah I see, made the changes in the latest commit. Thanks once again for your help, it's genuinely appreciated.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -46,7 +46,7 @@ sources:
 
   query: |
 
-    LET OneNoteMagic = 'e4 52 5c 7b 8c d8 a7 4d ae b1 53 78 d0 29 96 d3'
+    LET OneNoteMagic <= unhex(string='e4525c7b8cd8a74daeb15378d02996d3')
```
SwiftSecur, line 49: added the unhex in here.

content/exchange/artifacts/Windows.Forensics.OneNoteHunt.yaml
```diff
@@ -0,0 +1,203 @@
...
+      read_file(filename=OSPath, length=MaxRegexReadBytes) =~ CombinedRegex AS RegexMatched
+    FROM glob(globs=TargetGlob)
+    WHERE NOT IsDir
+      AND Size <= MaxFileSize
+      AND read_file(filename=OSPath, length=16) = OneNoteMagic
+
+    LET checked_files = SELECT *,
+      len(list={
+        SELECT String.Name
+        FROM yara(files=OSPath, rules=ImageRule, number=9999)
```
scudette: do you really need to produce 10k hits? you just care if one hit exists.

Because this is a subquery it will expand the entire yara plugin output into memory - potentially 10k element list

SwiftSecur: Fair point. Thanks once again. Ill make the change.

PR #1243, opened 2026-05-04
Add windows wazuh agent artifact

julianghill:
Adds a Windows Velociraptor artifact to install and configure the Wazuh agent.

I like to roll out Velociraptor during IR and being able to also install Wazhuh from there would be great.

The artifact supports installing the latest Wazuh Windows agent by resolving the GitHub release/package URL, or pinning a specific agent version with WazuhVersion, which is useful if the host has an already older version of Wazuh running. It also allows overriding the MSI URL directly and optionally validating the downloaded MSI with a SHA256 hash.

- Installs the Wazuh Windows agent silently via msiexec
- Configure Wazuh manager IP
- Supports optional RegistrationServer, RegistrationPassword, AgentName, and AgentGroup
- Supports comma-separated manager addresses
- Allows setting of a specific version
- Allows MsiUrl override in case own hosted
- Optionally updates existing ossec.conf
- Starts or restarts WazuhSvc
- Uploads the MSI install log for troubleshooting
- Tested against a Windows Velociraptor client and Wazuh manager
- Confirmed the Wazuh agent installed, connected, and appeared in the Wazuh dashboard

Feedback welcome!

### DIFF
--- a/content/exchange/artifacts/Windows.EventLogs.Hayabusa.yaml
+++ b/content/exchange/artifacts/Windows.EventLogs.Hayabusa.yaml
@@ -85,10 +85,9 @@ parameters:
    description: "Scan only common Event IDs for quicker scans"
    type: bool
    default: N
- - name: AddHostnameToFilename
-   description: "If checked, prepends the Hostname to the uploaded file."
-   type: bool
-   default: N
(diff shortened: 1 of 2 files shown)


julianghill:
Made a mistake.

PR #1244, opened 2026-05-04
Add Windows Wazuh agent install artifact

julianghill:
Windows Velociraptor artifact to install and configure the Wazuh agent.

I like to roll out Velociraptor during IR and being able to also install Wazhuh from there would be great.

The artifact supports installing the latest Wazuh Windows agent by resolving the GitHub release/package URL, or pinning a specific agent version with WazuhVersion, which is useful if the host has an already older version of Wazuh running. It also allows overriding the MSI URL directly and optionally validating the downloaded MSI with a SHA256 hash.

- Installs the Wazuh Windows agent silently via msiexec
- Configure Wazuh manager IP
- Supports optional RegistrationServer, RegistrationPassword, AgentName, and AgentGroup
- Supports comma-separated manager addresses
- Allows setting of a specific version
- Allows MsiUrl override in case own hosted
- Optionally updates existing ossec.conf
- Starts or restarts WazuhSvc
- Uploads the MSI install log for troubleshooting
- Tested against a Windows Velociraptor client and Wazuh manager
- Confirmed the Wazuh agent installed, connected, and appeared in the Wazuh dashboard

Feedback welcome!

### DIFF
--- a/content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
+++ b/content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
@@ -1,0 +1,459 @@
+name: Windows.Install.Wazuh.Agent
+description: |
+  Installs the Wazuh Windows agent and configures it to use the supplied
+  Wazuh manager IP address or hostname.
+
+  By default this artifact asks the GitHub releases API for the latest Wazuh
+  release. If that release exposes a Windows MSI asset, the artifact downloads
+  it directly. If not, it derives the official Wazuh packages URL from the
+  latest GitHub tag. Set WazuhVersion to pin to a manager-compatible agent
+  version, or set MsiUrl when you need to override the MSI download URL.
+
+  NOTE: Requires administrative privileges on the endpoint.
+
+author: Julian Hill - @julianghill
+type: CLIENT
+
+required_permissions:
+  - EXECVE
+  - FILESYSTEM_WRITE
+
+precondition: SELECT OS From info() where OS = 'windows'
+
+parameters:
+  - name: ManagerIP
+    description: Wazuh manager IP address or hostname. This is required.
+    default: ""
+  - name: RegistrationServer
+    description: Optional Wazuh registration server. If empty, the first ManagerIP address is used.
+    default: ""
+  - name: RegistrationPassword
+    description: Optional Wazuh enrollment password. If empty, WAZUH_REGISTRATION_PASSWORD is not passed to the installer.
+    default: ""
+  - name: AgentName
+    description: Optional Wazuh agent name. If empty, Wazuh uses its default agent naming behavior.
+    default: ""
+  - name: AgentGroup
+    description: Optional Wazuh agent group or comma-separated groups.
+    default: ""
+  - name: MsiUrl
+    description: Optional MSI URL override. If set, this takes precedence over WazuhVersion.
+    default: ""
+  - name: WazuhVersion
+    description: Optional Wazuh agent version to install, without the package release suffix. If empty, the artifact resolves the latest Wazuh release.
+    default: ""
+  - name: ExpectedSha256
+    description: Optional SHA256 expected hash for the downloaded MSI.
+    default: ""
+  - name: GithubReleaseApiUrl
+    description: GitHub API URL used to resolve the latest Wazuh release.
+    default: https://api.github.com/repos/wazuh/wazuh/releases/latest
+  - name: UpdateExistingConfig
+    type: bool
+    description: Update ossec.conf with ManagerIP after installation when possible. Supports comma-separated managers.
+    default: N
+  - name: WazuhConfigPath
+    description: Path to the Wazuh agent ossec.conf file.
+    default: C:\Program Files (x86)\ossec-agent\ossec.conf
+  - name: StartService
+    type: bool
+    description: Start or restart WazuhSvc after installation and configuration.
+    default: Y
+  - name: UploadInstallLog
+    type: bool
+    description: Upload the msiexec install log.
+    default: Y
+  - name: DownloadTimeoutSec
+    type: int
+    description: Download timeout in seconds.
+    default: 300
+
+sources:
+  - name: Install
+    query: |
+      LET TempDir <= tempdir()
+      LET MsiPath <= TempDir + "\\wazuh-agent.msi"
+      LET LogPath <= TempDir + "\\wazuh-agent-install.log"
+
+      LET EnvVars <= dict(
+        WAZUH_MANAGER=ManagerIP,
+        WAZUH_REGISTRATION_SERVER=RegistrationServer,
+        WAZUH_REGISTRATION_PASSWORD=RegistrationPassword,
+        WAZUH_AGENT_NAME=AgentName,
+        WAZUH_AGENT_GROUP=AgentGroup,
+        WAZUH_MSI_URL=MsiUrl,
+        WAZUH_VERSION=WazuhVersion,
+        WAZUH_EXPECTED_SHA256=ExpectedSha256,
+        WAZUH_GITHUB_RELEASE_API=GithubReleaseApiUrl,
+        WAZUH_UPDATE_EXISTING_CONFIG=if(condition=UpdateExistingConfig, then="true", else="false"),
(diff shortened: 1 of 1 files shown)


mgreen27:
> Windows Velociraptor artifact to install and configure the Wazuh agent.
> 
> I like to roll out Velociraptor during IR and being able to also install Wazhuh from there would be great.
> 
> The artifact supports installing the latest Wazuh Windows agent by resolving the GitHub release/package URL, or pinning a specific agent version with WazuhVersion, which is useful if the host has an already older version of Wazuh running. It also allows overriding the MSI URL directly and optionally validating the downloaded MSI with a SHA256 hash.
> 
> * Installs the Wazuh Windows agent silently via msiexec
> * Configure Wazuh manager IP
> * Supports optional RegistrationServer, RegistrationPassword, AgentName, and AgentGroup
> * Supports comma-separated manager addresses
> * Allows setting of a specific version
> * Allows MsiUrl override in case own hosted
> * Optionally updates existing ossec.conf
> * Starts or restarts WazuhSvc
> * Uploads the MSI install log for troubleshooting
> * Tested against a Windows Velociraptor client and Wazuh manager
> * Confirmed the Wazuh agent installed, connected, and appeared in the Wazuh dashboard
> 
> Feedback welcome!

Hi Julian - I notice you used a powershell script here, have you considered using the Velociraptor tools infrastructure? 
You could add the MSI as a tool then let Velociraptor manage the download and versioning.

Here is an example artifact that uses sysmon https://docs.velociraptor.app/artifact_references/pages/windows.sysinternals.sysmoninstall

You will just might need to use the Artifact.Generic.Utils.FetchBinary(IsExecutable=FALSE,.... )

julianghill:
@mgreen27-r7 Thanks for the feedback. I just updated it. Let me know if this is up to standard

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
```diff
@@ -0,0 +1,369 @@
...
+        $agentGroup = (@(Get-CommaSeparatedValues -Name "AgentGroup" -Value $agentGroup) -join ",")
+      }
+
+      if (-not (Test-Path -LiteralPath $msiPath)) {
+        throw "WazuhAgentMSI tool was not fetched to '$msiPath'."
+      }
+
+      $actualSha256 = (Get-FileHash -Algorithm SHA256 -LiteralPath $msiPath).Hash.ToLowerInvariant()
+
+      $msiArguments = @(
```
scudette: you actually do not need the complicated powershell - VQL execve already escapes args properly so this escaping is not really necessary

We generally prefer to not launch powershell scripts if we can help it.

julianghill: Removed the Powershell and tested.

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
```diff
@@ -0,0 +1,188 @@
...
+        then={
+          SELECT *
+          FROM chain(
+            stop={
+              SELECT "stop" AS Action, Stdout, Stderr, ReturnCode, Complete
+              FROM execve(argv=["sc.exe", "stop", "WazuhSvc"], length=100000)
+            },
+            wait={
+              SELECT "wait" AS Action, Stdout, Stderr, ReturnCode, Complete
+              FROM execve(argv=["ping.exe", "-n", "6", "127.0.0.1"], length=100000)
```
scudette: Just add a sleep to the previous step:

```
           stop={
              SELECT "stop" AS Action, Stdout, Stderr, ReturnCode, Complete
              FROM execve(argv=["sc.exe", "stop", "WazuhSvc"], length=100000)
              WHERE sleep(time=6)
           },
```

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
```diff
@@ -0,0 +1,188 @@
...
+        source=Value,
+        re="^\\s+|\\s+$",
+        replace="")
+
+      LET ManagerAddresses <= SELECT Trim(Value=_value) AS Address
+      FROM foreach(row=split(string=ManagerIP, sep=","))
+      WHERE Address
+
+      LET RegistrationServerAddresses <= SELECT Trim(Value=_value) AS Address
+      FROM foreach(row=split(string=RegistrationServer, sep=","))
```
scudette: You can specify the parameter type as CSV then add the default value like:

```yaml
    - name: RegistrationServer
      description: Optional Wazuh .... used.
      type: csv
      default:  |
         IP
         127.0.0.1
```

Then 
```
LET RegistrationServerAddresses <= RegistrationServer.IP
```

It is easier to use and also obvious so the user does not need to just know to add commas. Same goes for all the other parameters like that

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
```diff
@@ -0,0 +1,188 @@
...
+    description: Upload the msiexec install log.
+    default: Y
+
+sources:
+  - name: Install
+    query: |
+      LET TempDir <= tempdir()
+      LET LogPath <= TempDir + "\\wazuh-agent-install.log"
+
+      LET WazuhMSI <= SELECT FullPath, Name, DownloadStatus, Hash
```
scudette: Use = here instead of <= 

<= will materialize it always, but = will get it only on demand. If the parameters are wrong then it wont do anything

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
```diff
@@ -0,0 +1,188 @@
...
+        "msiexec.exe",
+        "/i",
+        WazuhMSI.FullPath[0],
+        "/qn",
+        "/norestart",
+        "/l*v",
+        LogPath,
+        "WAZUH_MANAGER=" + ManagerFinal,
+        "WAZUH_REGISTRATION_SERVER=" + RegistrationServerFinal,
+        if(condition=ReinstallExistingAgent, then="REINSTALL=ALL"),
```
scudette: you can use this syntax if you prefer

```
ReinstallExistingAgent && "REINSTALL=ALL",
...
RegistrationPassword && "WAZUH_REGISTRATION_PASSWORD=" + RegistrationPassword, 
```

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
julianghill: I applied this to WazuhMSI, but not to TempDir/LogPath because changing those two to = breaks the log path.
I tested that and hit msiexec error 1622: “Error opening installation log file.”

So this is intentionally kept:

LET TempDir <= tempdir()
LET LogPath <= TempDir + "\\wazuh-agent-install.log"

And the fetch was changed as suggested:

LET WazuhMSI = SELECT FullPath, Name, DownloadStatus, Hash

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
julianghill: Applied the CSV approach where the setting supports multiple values.

ManagerIP is now an IP column with a default value, so its clear I think that it has to be set.

AgentGroup is also now the csv type because it supports multiple groups. I left the default group empty, but added an empty quoted row so the UI shows an editable row by default which I think is nicer.
The VQL makes it so empty group rows are filtered before building WAZUH_AGENT_GROUP, so leaving that row blank does not pass an empty group to the installer.

I did not set RegistrationServer as csv because Wazuh enrollment uses a single registration server.

julianghill:
Thanks for the detailed feedback. Greatly appreciated.

content/exchange/artifacts/Windows.Install.Wazuh.Agent.yaml
scudette: So why split it here on , if there is only even one of them?

julianghill: Good point. I adjusted it to only trim the value entered. Decided against adding a saveguard to check if two command seperated values were added, because the user should know only one can be used.

PR #1253, opened 2026-05-12
Fix summaries for pages referenced by the 'children' shortcode

predictiple:
Fixes all instances where "description=true" is used, and that summary fields are available in addition to description fields.

Swapped order of preference in favour of Summary in the shortcode definition.


themes/hugo-theme-learn/layouts/shortcodes/children.html
```diff
@@ -55,10 +55,10 @@
 				{{end}}
 
 				{{if $.description}}
-					{{if .Description}}
-<p>{{.Description}}</p>
-					{{else}}
+					{{if .Summary}}
 <p>{{.Summary}}</p>
+					{{else}}
+<p>{{.Description}}</p>
```
scudette, line 59: Rather than change the theme lets just copy the file to the layouts

predictiple: Good point. Dunno why I didn't think of that

PR #1261, opened 2026-05-27
Add wrapper around flows() to handle new breaking change

misje:
These artifacts depends on the extra information that is now removed from default output from `flows()`.


scudette:
Maybe we should modify the new args to keep the default behavior.  The new behavior is just an optimization which would make it a lot faster but I think for people using the flows() plugin it probably wont matter too much.

scudette:
The main issue (which is the purpose of the original change) is that when someone launches a huge flow (with a large args like Yara rules) the doing `SELECT * FROM flows()` in the GUI will break the GUI due to large rows.

It is actually safe when running inside VQL (except slower and uses more memory) unless the large request is presented in the GUI.

misje:
Keeping the previous behaviour as default would not break any artifacts depending on the details, so that would probably be best.

scudette:
The flows plugin is now backwards compatible so this is not needed.

PR #1267, opened 2026-05-30
Add some best practices and advice for API client security

predictiple:


content/docs/deployment/security/_index.md
```diff
@@ -742,6 +742,119 @@ difficult for a compromised Velociraptor administrator account to
...
+If an API key is compromised, or suspected to be compromised, you
+don't need to struggle with complex certificate revocation lists
+(CRLs). Simply use the `acl grant` command to set an empty policy for
+that principal.
+
+```sh
+velociraptor acl grant --name "CompromisedKey" --role ""`
+```
+
+This immediately strips all permissions, rendering the certificate
```
scudette: This also means you have to reissue the key with a different name. That user name is effectively dead now and it is a security risk if it is recreated in future with the same name (because the old key will become valid again). Perhaps we need a proper CRL mechanism ?

content/docs/deployment/security/_index.md
```diff
@@ -742,6 +742,119 @@ difficult for a compromised Velociraptor administrator account to
...
+authentication methods you have set up for your users. The API
+client's authentication credentials are stored in a plain text file
+which needs to be tightly guarded, however the keys in such files can
+be passphrase-secured.
+
+API client certificates are **valid for one year**, by default. You
+must reissue them before they expire to avoid breaking your automation
+pipelines.
+
+**Encrypt Private Keys**: When generating an API config using
```
scudette, line 776: This is not really possible in most cases because a locked key needs to be interactively unlocked before use - this is not possible for automation scripts with no user access.

content/docs/deployment/security/_index.md
```diff
@@ -742,6 +742,119 @@ difficult for a compromised Velociraptor administrator account to
...
+`velociraptor config api_client`, use a password/passphrase to encrypt
+the private key. This ensures that if the config file is stolen, it
+cannot be used without the passphrase.
+
+**Offline CA**: For high-security environments, keep your CA private
+key offline. Only bring it online to sign new API certificates,
+preventing an attacker who gains root on the Velociraptor server from
+minting their own valid API keys.
```
scudette: you can sign and generate new key on a separate system - we dont need the entire gui/datastore to be actually enabled to do this so no need to bring the server online.

content/docs/deployment/security/_index.md
```diff
@@ -742,6 +742,119 @@ difficult for a compromised Velociraptor administrator account to
...
+
+Some permissions are more dangerous than they might appear because
+they provide a path to full server control:
+- `ARTIFACT_WRITER`: This role allows a user to edit artifact
+  definitions. Since VQL runs unrestricted on endpoints, an artifact
+  writer effectively has root access to every client on your network.
+- `EXECVE`: This permission allows the execution of arbitrary shell
+  commands on clients. It is typically reserved for administrators
+  because it can be easily misused.
+- `FILESYSTEM_WRITE`: On the server side, this allows creating files,
```
scudette, line 861: with caveates - https://docs.velociraptor.app/docs/deployment/security/#accessing-files-from-vql

content/docs/deployment/security/_index.md
predictiple: I've clarified that disabling access is just a stop-gap emergency measure. Deleting the user is probably a safer option but the only way to do that currently is via VQL.

> ```sh
> velociraptor acl grant --name "CompromisedKey" --role ""
> ```
> 
> This immediately strips all permissions, thus denying that user the
> ability to make an API calls. Any applications using that API user
> will need to be configured to use a new account with a new key. 
> 
> API client accounts can alternatively have their roles and permissions
> revoked via the User Management screen in the GUI.
> 
> Be aware that allowing the user account to exist with no roles
> assigned is still a risky situation as someone could accidentally
> grant permissions to it.
> 
> To permanently delete the user you can run the
> [user_delete()](/vql_reference/server/user_delete/)
> function in a notebook. There is currently no GUI option to delete a
> user.

content/docs/deployment/security/_index.md
predictiple: Now reads:

> `velociraptor config api_client` it is possible to set a
> password/passphrase to encrypt the private key. This ensures that if
> the config file is stolen, it cannot be used without the passphrase.
> However, this is not useful for credentials used by automated API
> clients because unlocking the key is a user-interactive process. For
> interactive API access (for example when using the API client that's
> included in the Velociraptor binary) where the API config might reside
> on a laptop, we recommend that you secure the key this way.

content/docs/deployment/security/_index.md
predictiple: Now reads:

> **Offline CA**: For high-security environments, keep your CA private
> key offline, which usually means keeping an offline copy of the full
> config with the CA key included in it, while removing the CA key from
> the config on your production server. This prevents an attacker who
> gains root on the Velociraptor server from being able to mint their
> own valid API keys. You can issue new API keys if you have the full
> offline config - access to the production server is not required,
> although you will also need to provision the new user on the server
> and grant it access.

content/docs/deployment/security/_index.md
predictiple: Thanks, I've added that link:

> - `EXECVE`: This permission allows the execution of arbitrary shell
>   commands on clients or the server (subject to
>   [additional restrictions](/docs/deployment/security/#accessing-files-from-vql)
>   that are in place by default). It is typically reserved for
>   administrators because it can be easily misused.

PR #1275, opened 2026-06-12
Create Custom.Windows.ADCS.CertificateAuthority.EDB.yaml

sec-pc:
Parses the Active Directory Certificate Services (ADCS) ESE database directly from disk. Reads the CA database at C:\Windows\System32\CertLog\\*.edb using the ntfs accessor to handle the file lock while CertSvc is running. Returns all certificate requests with key forensic fields including disposition (issued/revoked/failed/pending), requester and caller identity, subject DN, template, timestamps, and serial number. Joins the Requests, Certificates, and RequestAttributes tables on RequestID to produce a single complete row per certificate.


scudette:
Couple of issues:
1. Rename the artifact name so it does not have the Custom. prefix.
2. the tests fail due to spelling errors in the description - maybe add to the ignore word list so it will pass

Also a suggestion instead of the wide use of get(...) you might consider:

```sql
LET S = scope()

SELECT ... S.`$DistinguishedName`   AS DistinguishedName,
```

it might be a bit more readable

sec-pc:
Renamed the Artifact and added the words triggering the workflow fail were added to the ignore list.

scudette:
Thanks!

PR #1287, opened 2026-07-18
Add Journal.Fields artifact

misje:
Starting with the original, now abandoned, native parser of the systemd journal in VQL's binary parser (i.e. the native VQL binary parsing attempt, now moved into go), I saw a need to extract the indexes. Initially I just wanted to list boots. Then, as I dug deeper into the journal format, I decided to make the in-VQL parser more general.

Initial goal: Simply get a list of boots.

Since I realised that

- Listing all fields
- Listing unique field values, along with their earliest and latest journal entries

has real forensical value. This artifact provides both, and you specify which (or both) you want.

The collection is very fast and quickly retrieves a lot of unique values for a lot of journal keys, like _SYSTEMD and _COMM/_EXE.

I do not mind at all putting in the extra work needed to get this into the main project. As long as there is enough data in the test journal files.


misje:
Closing in favour of Velocidex/velociraptor#4933.

PR #1292, opened 2026-08-02
Server-side Sigma/Hayabusa analysis of collected EVTX (VQL-native successor to #1064)

ecapuano:
This PR adds server-side analysis of already-collected Windows event logs using Velociraptor's built-in Sigma engine and the curated Hayabusa ruleset — and adds a risk warning to the legacy endpoint-binary artifact.

It resurrects the intent of #1064 (credit to @shortstack for the original design), rebuilt to address each reason that PR was closed:

| #1064 close reason | This PR |
|---|---|
| Raw datastore paths break on 0.75+ (compressed datastore) | Files are staged with `copy(accessor="fs")`, which reads through the file store API and decompresses transparently. No raw paths, no external binary touching the datastore. |
| Should use `uploads()` / `file_store()` instead of path reconstruction | Flow uploads are enumerated with `uploads()`; nothing is reconstructed. |
| "Sigma processing is already built into velociraptor with the Hayabusa ruleset available at https://sigma.velocidex.com" | That is exactly the engine used. No Hayabusa binary is deployed to the endpoint **or** the server — the artifact delegates to the curated `Windows.Hayabusa.Rules` artifact via a runtime `collect()` call, so rules stay current with the curated pack and are never duplicated. |

## What's included

**`Server.Analysis.Windows.EventLogs.Hayabusa`** (SERVER) — takes a `ClientId`/`FlowId` of any completed collection that uploaded `.evtx` files (`Windows.Triage.Targets`, `Windows.KapeFiles.Targets`, etc.), stages the EVTX from the file store into a `tempdir()`, and evaluates the curated Hayabusa ruleset over it. Pass-through tuning: `RuleLevel` (all five upstream tiers), `RuleStatus`, `RuleTitleFilter`, `DateAfter`/`DateBefore`. Results carry `ClientId`/`FlowId`/`Fqdn` context.

Robustness behaviors:
- If the same channel file was uploaded more than once (e.g. volume shadow copies), the live-volume copy is preferred and skipped duplicates are logged.
- Missing curated package, missing `ClientId`/`FlowId`, or a flow with no matching EVTX each log an actionable message and return zero rows rather than failing silently or running the ruleset against nothing.
- Uploads lacking an original endpoint path fall back to their file store filename rather than being silently excluded.

**`Server.Monitor.Windows.EventLogs.Hayabusa`** (SERVER_EVENT) — watches `System.Flow.Completion` and automatically analyzes matching flows (default regex: `Windows\.KapeFiles\.Targets|Windows\.Triage\.Targets`). Results appear under Server Events. Analyses run inline and serialized; this is documented as a deliberate simplicity tradeoff.

**`Windows.EventLogs.Hayabusa`** (existing) — description now carries a warning after the summary paragraph: the deployed binary is commonly flagged/quarantined by AV/EDR, `UpdateRules` makes the endpoint reach out to GitHub, and on-endpoint analysis modifies forensic state — with a pointer to the native alternative. Functionality unchanged.

## Why server-side

Deploying `hayabusa.exe` to endpoints routinely trips third-party security tooling (alerting defenders, quarantining the tool mid-collection) and pollutes endpoint forensic state. Collecting raw EVTX and analyzing centrally avoids all of that, works retroactively on collections already sitting on the server, and needs no per-endpoint tool staging.

## Tested

End-to-end on a Windows Server 2022 VM running Velociraptor **0.77.1** in GUI mode:

- Manual run against a 336-file (~40MB) `Windows.Triage.Targets` EventLogs collection: **604 detections in ~14 seconds**, including `Potentially Malicious PwSh`, `Windows Defender Threat Detected`, `Permanent WMI Event Consumer`.
- Monitor auto-triggered on a fresh flow completion and produced 645 rows into Server Events with no operator action; server logs show the `uploads() → copy(fs://…) → sigma` pipeline executing.
- After a review hardening pass, the full suite was re-run: identical 604-row baseline (no behavioral regression), the `Critical, High, Medium, and Low` tier filters correctly, and all failure-path guards produce clean log messages with zero rows.
- `velociraptor-0.77.1 -v artifacts verify ./content/exchange/artifacts/*yaml` passes (exit 0) with these files included. The dynamic `collect()` invocation is deliberate: a static `Artifact.Windows.Hayabusa.Rules` reference fails `artifacts verify` since the curated artifact is not built-in.

## Notes

- Prerequisite (documented in all descriptions): the curated Sigma package must be imported so `Windows.Hayabusa.Rules` resolves — `Server.Import.CuratedSigma` on 0.74, `Server.Import.Extras` on 0.75+.
- `required_permissions: [SERVER_ADMIN]` (the `fs` accessor requires it).
- Temp disk usage ≈ decompressed size of the collected EVTX set per analysis; the tempdir is removed when the collection completes.
- Only standard-named channel files under the staging directory are scanned (a constraint of the `Windows.Sigma.Base` log-source model).

### DIFF
--- a/content/exchange/artifacts/Server.Analysis.Windows.EventLogs.Hayabusa.yaml
+++ b/content/exchange/artifacts/Server.Analysis.Windows.EventLogs.Hayabusa.yaml
@@ -1,0 +1,180 @@
+name: Server.Analysis.Windows.EventLogs.Hayabusa
+author: Whitney Champion - bluesky @whit.zip, Eric Capuano - bluesky @eric.zip
+description: |
+  Runs the curated Hayabusa Sigma ruleset against Windows event logs
+  that were already uploaded to the server by a completed collection
+  flow, using Velociraptor's built-in Sigma engine. No Hayabusa binary
+  is deployed to the endpoint or the server - the entire analysis runs
+  in VQL on the server side.
+
+  This artifact is intended to post-process collections that upload
+  `.evtx` files, such as `Windows.KapeFiles.Targets` or
+  `Windows.Triage.Targets` (or any other collection that uploads raw
+  event logs). It copies the collected EVTX files out of the server's
+  file store into a temporary directory, then evaluates the
+  `Windows.Hayabusa.Rules` artifact over that directory.
+
+  **Prerequisite**: The curated Hayabusa Sigma package must be imported
+  into the server first so that the `Windows.Hayabusa.Rules` artifact
+  resolves. Import it using the built-in `Server.Import.CuratedSigma`
+  artifact on Velociraptor 0.74, or `Server.Import.Extras` on 0.75 and
+  later. The curated packages are published at
+  https://sigma.velocidex.com/. If the package is missing at runtime
+  this artifact logs an error and returns no rows.
+
+  This artifact can be called from within another artifact:
+
+    `SELECT * FROM Artifact.Server.Analysis.Windows.EventLogs.Hayabusa(ClientId=..., FlowId=...)`
+
+  Notes:
+
+  - Requires Velociraptor 0.74 or later.
+  - The analysis needs temporary disk space on the server roughly equal
+    to the decompressed size of the collected EVTX set. The temporary
+    directory is removed automatically when the collection completes.
+  - Uploaded files are flattened to their basename in the temporary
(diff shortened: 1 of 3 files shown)


scudette:
This looks great! Thanks!

PR #1298, opened 2026-09-01
Migrate backup/restore KB article into a proper deployment documentation page

predictiple:
Also clarified some explanations and verified that it still accurately describes the automated daily backup features and behaviour.


content/docs/deployment/backups/_index.md
```diff
@@ -0,0 +1,380 @@
...
+other org's backup, you can also use the `prefix` parameter to restore
+only a specific org's data from the combined file.
+
+Note that ACL records are not automatically restored for security
+reasons. However you can restore them from the backup data in a
+Velociraptor notebook, after carefully reviewing the data. For
+example, this VQL would restore the users and ACLs for a specific org:
+
+```vql
+SELECT *
```
scudette: We probably could refactor this abit
```sql
SELECT
    user_create(
      user=Principal.name,
      orgs=Principal.orgs[0].id,
      roles=Policy.roles),
    user_grant(
      user=Principal.name,
      orgs=Principal.orgs[0].id,
      policy=Policy)
FROM parse_jsonl(
      filename="/tmp/extracted_backups/orgs/O123/acls.json")
```

content/docs/deployment/backups/_index.md
```diff
@@ -0,0 +1,380 @@
...
+The backup package contains data from various
+`Providers`. Each provider is responsible for saving some aspect of
+the server's configuration. To keep the size down, providers do not
+save bulky items like collected data, rather only metadata is stored
+about the server.
+
+For example, the hunts are saved, but not the list of clients that
+have already provided results to the hunt (since the associated
+collection data is not backed up). This means that when restoring the
+backup on a new server, clients will participate in existing hunts
```
scudette, line 161: Maybe we should restore the hunts into the paused state so the user can determine if the hunt should be recollected.

content/docs/deployment/backups/_index.md
```diff
@@ -0,0 +1,380 @@
...
+```vql
+SELECT Name FROM glob(globs="/backups/*", accessor="fs")
+ORDER BY Name DESC LIMIT 1
+```
+
+No service restart is required after restoring a backup.
+
+{{% notice warning "Restoring rolls back server state" %}}
+
+Restoring a backup rolls the server state back to what it was when the
```
scudette: I am not sure this is correct? do we actually delete clients and flows that were added since the backup? 

Should we?

content/docs/deployment/backups/_index.md
```diff
@@ -0,0 +1,380 @@
...
+backup was taken. Any changes made since then - such as new clients,
+hunts, or users - are lost.
+
+{{% /notice %}}
+
+{{% notice note "The server configuration is not part of the backup" %}}
```
scudette: We currently do not back up the configuration because we dont want the sensitive keys to be copied multiple times into backup files - I feel this might increase the risk of compromise if there are multiple copies of it.

content/docs/deployment/backups/_index.md
predictiple: I can't think of a simpler alternative. Can you suggest how? It can be refactored to use a custom function but that's less simple. The idea here was a 1-liner that can be pasted into a notebook.

In my mind the the broader issue with it is that backup_restore() won't restore the ACLs, but someone can run a query that does, which can be made into a server artifact that does... in which case why doesn't backup_restore() just do it in the first place? So this query seems like a workaround for an obstacle that exists by design.

I suspect the reasoning for not restoring ACLs is that people could accidentally restore accounts that were deleted/disabled. But generally you would restore from the latest backup which is going to be a day old, and therefore deleted accounts will be recent occurences and easy to track.

content/docs/deployment/backups/_index.md
predictiple: Yes that seems like it would be a good safety catch.

content/docs/deployment/backups/_index.md
predictiple: It is actually partially correct: client info gets wiped before being rewritten, but hunts, notebooks and artifacts are added to the existing ones.

But obviously changes since the time of the backup do get lost, and that's really what that statement is referring to. I'll reword it.

predictiple: New wording in https://github.com/Velocidex/velociraptor-docs/pull/1298/changes/bf058e0b762679e8cf17d2e2f5c01082fe8a7196

content/docs/deployment/backups/_index.md
predictiple: That makes sense to me. The reason I added the note is that it might be surprising to users that the config isn't included in the backup. On Discord we've had questions along the lines of "is the backup sufficient to restore my server?". This just points out that backups of the config file are a separate thing to consider.

content/docs/deployment/backups/_index.md
predictiple: Maybe ACLs should be restorable via backup_restore() but have a disabled-by-default safety behaviour?

scudette: Is the VQL I suggested above not equivalent to what you have ?

scudette: Maybe we need a more complete restore options dict which controls many aspects of the restore (including which providers, ACLs etc).

content/docs/deployment/backups/_index.md
scudette: This is interesting! the backup replaces the client index but the client records are not actually removed. This means that data will linger on disk without being deleted. If new client is still connected - looks like the key will be there so it will no re-enrol immediately but may re-enrol later. Then the client's flows etc will be visible. It also means that client data will not be deletable while it is not visible in the client index.

Maybe we need to keep the new client records in the index as well.

content/docs/deployment/backups/_index.md
predictiple: Ah sorry, I thought you were just quoting the original query. Your refactor is better. Added in https://github.com/Velocidex/velociraptor-docs/pull/1298/changes/dee7ec724b61d1807607b442a43123dd7ee8f0b4

Yes, most traditional system restore utilities have some form of wizard with low-risk options selected by default and higher-risk ones as optional selections (usually with "are you really sure?" warnings displayed if you select them).

PR #1303, opened 2026-09-14
MacOS.Applications.SavedState: reach the macOS 15+ daemon container path

### LINKED ISSUE #1300: MacOS.Applications.SavedState: add the macOS 15+ Daemon Containers path, and state the decryption facts in the description
Proposed, from decoding the format on macOS 26 (Terminal.app). I have not yet written or run the VQL below, so I am posting the analysis first in case the approach is wrong; happy to turn it into a PR if it looks right.

1. `SavedStateGlob` is a single comma-separated parameter, so the default could become `/Users/*/Library/Saved Application State/com.apple.**,/Users/*/Library/Daemon Containers/*/Data/Library/Saved Application State/*.savedState/**`. Since macOS 15 the per-app directories live under the talagentd daemon container and the first glob alone returns nothing there, and reading the container path needs Full Disk Access, so a collection without it will still come back empty. One consequence the query has to handle: `Name` is currently `OSPath[4]`, which is the `.savedState` directory under the old path but the container uuid under the new one, because there the `.savedState` component sits at `OSPath[8]` (Users, user, Library, Daemon Containers, container uuid, Data, Library, Saved Application State, then the `.savedState` directory) and is an app uuid rather than a bundle id. `_User` at `OSPath[1]` is unaffected and works under both. The uuid resolves through `ApplicationMapping.plist` in the same `Saved Application State` directory (a flat array alternating a dictionary carrying `protected.signingIdentifier` with the uuid string that follows it); a resolver would read that plist to map uuid to bundle id and fall back to the uuid when there is no entry.

2. Description only (the artifact reads plist metadata and does not decrypt; that would stay unchanged): records in `data.data` are `NSCR1000`, a big-endian 32-bit window id, a big-endian 32-bit total length that includes the 16-byte header, then AES-128-CBC ciphertext keyed by the matching window's 16-byte `NSDataKey` from `windows.plist`, with a zero IV and no MAC; the file is append-only, and current state is the last record for each (window id, `NSUserInterfaceItemIdentifier`) pair; `window_N.data` files are encrypted window bitmaps under a separate keychain key, not `NSDataKey`.

Plus a `reference:` entry for `https://www.mothersruin.com/software/Archaeology/reverse/appstate.html`, the Mothers Ruin write-up that documents most of this layout independently.

willmcginnis:
Fixes #1300. On macOS 15 and later this artifact returns nothing, because the per-application saved state directories moved under the `talagentd` daemon container and the default glob does not reach them. I opened that issue with the analysis and offered a PR; there was no reply either way, so here it is rather than leaving the offer hanging. Happy to close it if the approach is wrong.

## What changes

1. `SavedStateGlob` gains the daemon container location. It is already a comma separated parameter, so the legacy entry is untouched and both locations are globbed.
2. `Name` is derived from the right component for each layout. Under the legacy path the `.savedState` directory is `OSPath[4]`; under the daemon container path that component is the container uuid and `.savedState` sits at `OSPath[8]` (`Users`, user, `Library`, `Daemon Containers`, container uuid, `Data`, `Library`, `Saved Application State`, then the directory). Without this, `Name` comes back as a container uuid. `_User` is `OSPath[1]` and is correct under both, so it is untouched.
3. Description gains the macOS 15 path change, the Full Disk Access requirement, and the `data.data` record format for anyone decrypting outside this artifact.
4. Adds the Mothers Ruin write up as a reference, which documents most of the format independently.

Two deliberate choices worth flagging, because both are visible in the diff:

**Granularity is kept identical to the legacy glob.** The new entry is `*.savedState`, not `*.savedState/**`, so it returns one row per application state directory exactly as `com.apple.**` does today. I had `/**` at first and it is wrong here: a trailing `/**` returns the descendants and excludes the directory, which drops the `Btime` and `ModTime` rows the description calls the primary signal.

**The new entry cannot be scoped to a vendor prefix.** Under the daemon container path the directories are named with an application uuid rather than a bundle id, so there is no `com.apple.` to match on. That entry is therefore `*.savedState` and will return non-Apple applications where the legacy entry would not. It is an asymmetry forced by the path format rather than a choice, but it is a real behaviour difference and I would rather point at it than have you find it.

The `regex_replace` pattern and the `else` branch are textually unchanged, so the legacy `Name` derivation is the same expression it was before.

## Tested

macOS 26.6.2 (build 25G83, arm64) virtual machine, velociraptor 0.77.2.

| artifact | exit | rows |
|---|---|---|
| current `master` | 0 | **0** |
| this PR | 0 | **8** |

```
velociraptor --definitions <dir> artifacts collect MacOS.Applications.SavedState --format json
```

The current version exits cleanly with empty stdout and empty stderr, which is what makes this easy to miss: it reads as a host with no saved state rather than as a glob that cannot reach it.

Positive control, so that the zero is a real reading rather than a broken query:

```
velociraptor query "SELECT OSPath FROM glob(globs='/Users/*/Library/Daemon Containers/*/Data/Library/Saved Application State/*.savedState')"
```

returns the `.savedState` directories on that host, and the host has nothing at all under the legacy path. So the data exists and only the default glob was missing it.

Those 8 rows are **4 distinct application state directories** (Safari, Terminal, Finder and System Settings), each with `Name`, `_User`, `Btime` and `ModTime` populated and no null `Name`. They appear twice because that base image symlinks a second user home to the first, so the `/Users/*` glob legitimately matches both. `velociraptor artifacts verify`, which is what the repo CI runs, returns 0 on the patched file.

Two limits I would rather state than paper over. That host has no data under the legacy path, so I could not exercise the old branch end to end; the claim above is textual, not behavioural. And the collection ran as root, which is not the same as establishing Full Disk Access, so treat the FDA line in the description as the documented requirement rather than something this test isolated. "Since macOS 15" is likewise from the referenced write up and the path layout, not from a macOS 15 host I tested.

## Two things I noticed that are not in this PR

Both are pre-existing and I have deliberately left them alone rather than bundling them.

`**` only means "recurse" when it is a whole path component. Embedded as `com.apple.**` it behaves like `*`, so the legacy glob returns the `.savedState` directory and never `data.data` or `windows.plist`. That means `OSPath =~ "windows.plist"` can never be true today, and `DockMenuName`, `WindowTitle` and `_WindowDetails` are therefore always null on every host. If that is not intended it is a small change, but it widens what the artifact collects and felt like your call rather than mine.

Related: collecting the files as well as the directory would make those columns work, and is what I originally wrote. I backed it out to keep this PR to "reach the new location" with no change in row shape. Happy to send it separately.

## The uuid mapping, documented but not resolved

Under the daemon container path `Name` is an application uuid, because that is what the directory is called. `ApplicationMapping.plist`, in the same `Saved Application State` directory, maps them: a flat array alternating a dictionary carrying `protected.signingIdentifier` with the uuid string that follows it. The description now documents this.

I did not put a resolver in this PR, to keep the diff small. I do have working VQL for it, verified on that host, which resolves the uuids above to `com.apple.Safari`, `com.apple.Terminal`, `com.apple.finder` and `com.apple.systempreferences`. It needs a second pass over the mapping plist and a computed index, so it is noticeably more code than the rest of this change. Happy to add it here or as a follow up, whichever you prefer.


scudette:
Sorry your previous suggestions went without response.  The main decision is if this artifact should be part of the sqlite hunter or a separate one. Since it is already in there we should probably fix it but long term maybe remove it to the sqlite hunter. 

you are right in that ** should probably never be used together with anything - it probably should be something like `com.apple.*/**/.savedState`.

Dont worry about keeping the PR small - if you have additional contribution to make it more useful we welcome those

willmcginnis:
Thanks, and the "fix it here now, maybe move it to the sqlite hunter later" call works for me. I've kept the new commit contained to the one `sources:` query and it adds no parameters, which should at least keep a later move small.

I've pushed a second commit that fixes three columns which come back null under the default globs.

**The defect.** `DockMenuName`, `WindowTitle` and `_WindowDetails` are each guarded by `if(condition = OSPath =~ "windows.plist", ...)`. Under the default globs the rows are `*.savedState` **directories**, so `OSPath` is never a file inside one and that condition does not hold, which leaves those three columns null. The reason is the same fact as the glob note below: `**` embedded in a component behaves like `*` and does not recurse. The description advertises `NSTitle`, `NSWindowID`, `NSDataKey` and the dock-menu name, and none of them comes back. (`SavedStateGlob` is a parameter, so someone pointing it at files directly is a different case. I mean the shipped defaults and the normal saved-state layout.)

**The fix** computes the file path from the directory row in the `LET` (`OSPath + "/windows.plist" AS _WindowsPlist`) and reads `plist(file=_WindowsPlist)`, instead of matching against `OSPath`. It also adds `NSWindowID` while it's there. I deliberately did not reach the files by glob, because that changes granularity from one-row-per-directory to one-row-per-file; this way the directory's `Btime`/`ModTime` stay on the row, and the description calls those the primary signal. (A glob-based version could presumably recover the parent's metadata too, so that is a property of this implementation rather than an argument that the other shape cannot work.)

**Two things I only found by running it:**

On 0.77.2, `plist()` unwraps a single-element array into the dict. So for an application with exactly one saved window, `plist(...)[0].NSTitle` returns null, while the same expression returns the first window's title where several are present. On a real macOS 26 host, 8 of 26 `windows.plist` files contain exactly one window, so indexing `[0]` would have been silently empty for roughly a third of them, with nothing to indicate a problem. `len()` is no help either: on the one-window case it returns the dict's key count rather than a window count. The visible consequence is that `WindowTitle` is a scalar where one window is present and a list where several are, and null where the file has no `NSTitle` at all (18 of those 26 files carry one). I've left that rather than picking one window and discarding the rest, but say the word if you would prefer it always be a list.

A missing `windows.plist` returns null rather than erroring, so no existence guard is needed. That matters more than it sounds: on the same host, 75 of 101 saved-state directories have no `windows.plist` at all, so the absent case is the common one rather than an edge case.

**On the glob**, agreed. `com.apple.*/**/.savedState` is the natural way to write it, but it happens to return nothing, because `.savedState` is a name *suffix* rather than a child directory: the directories are named `com.apple.TextEdit.savedState`, one component, so that pattern looks for a child literally named `.savedState`. Measured against a synthetic fixture, velociraptor 0.77.2 on Linux. I could not exercise this on macOS, though neither the glob nor `plist()` looks platform-dependent here:

```
com.apple.**                 -> 2 matches   (current)
com.apple.*/**/.savedState   -> 0 matches
com.apple.*.savedState       -> 2 matches
```

`com.apple.*.savedState` avoids `**` entirely, and is what I would suggest if you want it out of the default. It's a one-line parameter change and I'm happy to fold it into this PR. I left it out for now because it changes the default glob's behaviour rather than fixing a bug, so it seemed like your call rather than mine.

The macOS 15+ half of the glob cannot carry a `com.apple.` prefix either way, since those directories are UUID-named, which is what the first commit is about.

